Issue #13, opened 2018-01-08
Investigate how an full core file can be exported to Avro


fmendezh:
Create an Avro schema with all DwC, DC and Gbif terms: http://rs.tdwg.org/dwc/terms/ 
http://purl.org/dc/terms/

timrobertson100:
It might have been discussed, but I propose consider if you need to do this.  

The intention in [the readme](https://github.com/gbif/pipelines/blob/master/README.md) was to keep verbatim data in a star schema form which is modelled [in this avro schema](https://github.com/gbif/pipelines/blob/master/core/src/main/avro/dwca.avsc).  

As categories of data are interpreted schemas be created for the interpreted data, so you end up with e.g.:

 1. `verbatim.avro` holding the raw data
 2. `temporal.avro` holding the interpreted date/time information
 3. `gbif-taxonomy.avro` holding the interpreted taxonomy according to the GBIF backbone

It does mean that to access data from the verbatim data you would need a function like `getString(Term field)` (e.g. `getString(DwC.scientificName`) rather than `getScientificName()` but I believe this will prove better as more and more data are added since the verbatim schema remains unchanged.

This design should allow intuitive extensibility (e.g. adding things like interpretation for protected areas, sites of scientific interest etc)  and only storing information that are interpreted into the new files.

fmendezh:
@timrobertson100 for your surprise @aalbatross had the same idea: a generic dwca Avro schema and specific Avro schemas for each group of transformations/interpretations

PR #17, opened 2018-01-16
Feature/dw ca avro schema generation

aalbatross:
1.A project to generate DwCA related terms in avro
2. putting the generated schema in core project.


DwCASchemaGenerator/pom.xml
```diff
@@ -0,0 +1,72 @@
...
+        <configuration>
+          <source>1.8</source>
+          <target>1.8</target>
+        </configuration>
+      </plugin>
+    </plugins>
+    <!-- </pluginManagement> -->
+  </build>
+
+  <dependencies>
```
fmendezh: all dependencies versions must be moved to a properties section for better management and keep consistency with other gbif projects

DwCASchemaGenerator/pom.xml
```diff
@@ -0,0 +1,72 @@
...
+      <artifactId>avro</artifactId>
+      <version>1.8.2</version>
+    </dependency>
+    <dependency>
+      <groupId>org.jsoup</groupId>
+      <artifactId>jsoup</artifactId>
+      <version>1.8.3</version>
+    </dependency>
+    <dependency>
+      <groupId>org.testng</groupId>
```
fmendezh: better if we stick to use junit by now

cgendreau:
Just a quick note and please remember I don't have all the information.

My first understanding was that this was not necessary (see https://github.com/gbif/pipelines/issues/13). If that changed please let me (us) know.

If it is now required (which I assume it is) and it's there to stay, I would suggest to parse to the XML instead of the HTML page. @mdoering would now better what is the better source to use but I know a XML is available on GitHub (https://github.com/tdwg/dwc/blob/master/standard/documents/xml/tdwg_dwcterms.xsd).

timrobertson100:
XML versions are on http://rs.gbif.org/core/

DwCASchemaGenerator/src/main/java/org/gbif/DwCAAvroSchemaGenerator.java
```diff
@@ -0,0 +1,48 @@
+package org.gbif;
```
fmendezh: This project doesn't follow the name convention for Maven artifacts, a better approach is to generate  Avro schemas form XSD or XML files, there are several libraries for that, take for example: https://github.com/elodina/xml-avro

DwCASchemaGenerator/src/main/java/org/gbif/dwca/generator/util/SchemaType.java
```diff
@@ -0,0 +1,10 @@
+package org.gbif.dwca.generator.util;
+
+/**
+ * Type of schema
+ *
+ * @author clf358
```
fmendezh: we discourage the use of @autor

Issue #10, opened 2018-01-08
Validate uniqueness of a field (occurrenceId)


fmendezh:
All the occurrence id fields must be have unique value and can't be null

PR #22, opened 2018-01-19
Validate uniqueness of a field (occurrenceId) #10

muttcg:


demo/src/main/java/org/gbif/pipelines/demo/transformation/validator/UniqueOccurrenceIdTransformation.java
```diff
@@ -0,0 +1,74 @@
...
+import org.apache.beam.sdk.values.TypeDescriptor;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Transformation for filtering all duplicate records with the same occurrenceId
+ *
+ * TODO: Probably it should not be UntypedOccurrenceLowerCase class, some common class with occurrenceID as a key.
+ */
+public class UniqueOccurrenceIdTransformation extends ValidatorsTransformation<UntypedOccurrenceLowerCase> {
```
fmendezh: apply this to the ExtendedRecord raw class

Issue #18, opened 2018-01-19
Interpret scientific name and taxonomy fields using the new name matching service


asturcon:
current interpreter: https://github.com/gbif/occurrence/blob/master/occurrence-processor/src/main/java/org/gbif/occurrence/processor/interpreting/TaxonomyInterpreter.java

**Subtasks:**
- [x] Basic interpreter and structure
- [x] Move taxonomy interpretation to gbif module
- [x] Use new FP classes in taxonomy interpretation

asturcon:
Diagram with the taxonomic interpretation implemented in this issue:
![taxonomy interpretation diagram](https://user-images.githubusercontent.com/25691745/35799176-3d7ad938-0a65-11e8-9943-0886ca26879b.png)

@fmendezh should we do something else besides adding an Issue when the match type is NONE? In the old interpretation they set the kingdom to unknown, but we're following a different approach since we just serialize the WS response to an Avro file. I'm not sure about the implications.

I also add a diagram with the rank interpretation that I used in development to understand the process (it's the same interpretation as the old one):
![interpret rank diagram](https://user-images.githubusercontent.com/25691745/35798217-638c0320-0a62-11e8-8345-a61099b594a4.png)

PR #29, opened 2018-02-02
#18 interpret taxonomy fields

asturcon:
#18 There are some ToDo to check but can be done later, so we don't delay the integrate of this.

Also pending to finish the diagrams.


core/src/main/java/org/gbif/pipelines/core/config/TargetPath.java
```diff
@@ -11,20 +11,27 @@
...
     directory = dir;
     this.fileName = fileName;
   }
 
+  /**
+   * Needed for json serialization with Jackson. Do NOT invoke it programmatically!! Setters are not created to
+   * avoid the use of this constructor.
+   */
+  public TargetPath() {}
+
```
asturcon: @fmendezh these are the changes for the serialization

core/src/main/java/org/gbif/pipelines/core/config/TargetPath.java
```diff
@@ -34,10 +41,9 @@ public String getFileName() {
   }
 
   /**
-   *
    * @return a full path which concatenates the directory and file names.
    */
-  public String getFilePath(){
+  public String filePath() {
```
asturcon: needed for serialization, not to be interpreted as a property

asturcon:
closes #32

asturcon:
@fmendezh @muttcg I think it's ready for another review

Issue #73, opened 2018-03-16
Modify DwCAReader to support expanded/uncompressed DwCA files


fmendezh:
https://github.com/gbif/pipelines/blob/master/core/src/main/java/org/gbif/pipelines/core/io/DwCAReader.java expands/uncompresses a DwCA to read its content, this step can be skiped if a another process extracted the archive content  in a directory; therefore, this class has to be modified to support reading from an archive and also from directory that contains a expanded DwCA.

PR #74, opened 2018-03-19
#73 Modify DwCAReader to support expanded/uncompressed DwCA files

muttcg:


core/src/main/java/org/gbif/pipelines/core/io/DwCAReader.java
```diff
@@ -21,13 +21,19 @@
  */
 public class DwCAReader {
```
fmendezh: add javadoc to both constructors to make it clear when one should be used instead the other

PR #77, opened 2018-03-20
#69 duplicate xml records detection

### LINKED ISSUE #69: Duplicate XML records detection
Due to the nature of crawling the XML files will hold duplicate records.  This process must remove the duplicates, using the same mechanism [as used currently](https://github.com/gbif/occurrence/blob/master/occurrence-parser/src/main/java/org/gbif/occurrence/parsing/xml/XmlFragmentParser.java#L150) - this will surely raise questions to be answered during implementation.  In removing duplicates we must recognise that we wish to run minimal memory requirements (100s MB only) and therefore need a mechanism to handle this without introducing a dependency on an external technology.  Something along the lines of [mapdb](http://www.mapdb.org/) which GBIF use in Checklist indexing to good effect would be sensible as an embedded option.

Sub tasks related:
- [x] Create a map cache with MapDB and test it
- [x] Set a unique ID to the records that don't have it using the holy triplet.

This task is part of epic #60 "ExtendedRecord builder for XML files"
This task depends on task #64 "Move occurrence-parsers to pipelines/xml-occurrence-parsers"

asturcon:
Observations when testing mapdb:
- When using the overflow it didn't work as expected since it wasn't keeping unique keys between the memory and the disk map, so this option was discarded.
- The mmap option gave the best performance (around 10-12 times faster). Tests were done in my machine and were not very strict in the measures. I also enabled the fileChannel to be used in case the mmap is not supported in the machine where the application is running.
- This configuration should be tested in integration tests to make sure that it doesn't raise any issue.

@muttcg check the changes I did to use the map and set the missing IDs. Maybe there is a better way to fit it into the existing code.

closes #69


xml-occurrence-parsers/src/main/java/org/gbif/xml/occurrence/parser/parsing/extendedrecord/ExtendedRecordConverter.java
```diff
@@ -22,10 +22,12 @@ private ExtendedRecordConverter() {
   public static ExtendedRecord from(RawOccurrenceRecord rawRecord) {
 
     String recordId = rawRecord.getId();
+    // FIXME: check if this has to be done here or in the XmlFragmentParser or in both places
     if (Strings.isNullOrEmpty(recordId)) {
```
fmendezh: can all this be removed since it is already done by XmlFragmenterParser

xml-occurrence-parsers/src/main/java/org/gbif/xml/occurrence/parser/parsing/xml/XmlFragmentParser.java
```diff
@@ -208,4 +213,34 @@ public static RawOccurrenceRecord parseRecord(byte[] xml, OccurrenceSchemaType s
...
+    if (records == null) {
+      return Collections.emptyList();
+    }
+
+    List<RawOccurrenceRecord> result = new ArrayList<>(records.size());
+    for (RawOccurrenceRecord record : records) {
+      // try to set an id in case it is missing
+      if (Strings.isNullOrEmpty(record.getId())) {
+        // try to create a triplet
+        Triplet triplet = null;
```
fmendezh: Triplet triplet = new Triplet(record.getInstitutionCode(), record.getCollectionCode(), record.getCatalogueNumber());
          record.setId(OccurrenceKeyHelper.toKey(triplet));


or

          record.setId(OccurrenceKeyHelper.toKey(new Triplet(record.getInstitutionCode(), record.getCollectionCode(), record.getCatalogueNumber())));

and remove L239?

xml-occurrence-parsers/src/main/java/org/gbif/xml/occurrence/parser/parsing/xml/XmlFragmentParser.java
```diff
@@ -208,4 +213,34 @@ public static RawOccurrenceRecord parseRecord(byte[] xml, OccurrenceSchemaType s
...
+        try {
+          triplet = new Triplet(record.getInstitutionCode(), record.getCollectionCode(), record.getCatalogueNumber());
+        } catch (Exception e) {
+          LOG.info("Could not create a triplet for record with institution code {}, collection code {} and catalogue "
```
fmendezh: in logback you can add a Throwable and it will be handled as the error context and is ignored as argument, this means that you can do:

LOG.info("Could not create a triplet for record with institution code {}, collection code {} and catalogue "
                   + "number {}", record.getInstitutionCode(), record.getCollectionCode(), record.getCatalogueNumber(), e);

See: https://github.com/qos-ch/logback/blob/master/logback-classic/src/main/java/ch/qos/logback/classic/spi/LoggingEvent.java#L127

xml-occurrence-parsers/src/main/java/org/gbif/xml/occurrence/parser/parsing/xml/XmlFragmentParser.java
```diff
@@ -208,4 +213,34 @@ public static RawOccurrenceRecord parseRecord(byte[] xml, OccurrenceSchemaType s
 
     return results;
   }
+
+  /**
+   * Filters the records by discarding the ones without ID.
+   */
+  private static List<RawOccurrenceRecord> filterRecords(List<RawOccurrenceRecord> records) {
```
fmendezh: filterRecords is not a good name, forceIdentifiers?? any other idea?

xml-occurrence-parsers/src/test/java/org/gbif/xml/occurrence/parser/parsing/validators/UniquenessValidatorTest.java
```diff
@@ -0,0 +1,89 @@
...
+
+  @Test
+  public void givenDuplicatedIdsWhenMappedThenDuplicateFound() {
+    // increase it to test with a higher volume of data
+    int n = 10;
+    int duplicatesFound = 0;
+
+    try (UniquenessValidator validator = UniquenessValidator.getNewInstance()) {
+      // add duplicates
+      List<String> duplicates = new ArrayList<>(Arrays.asList("12.1", "1234.1", "111.1", "12.12"));
```
fmendezh: new ArrayList<>(.. can be removed
and maybe something like this can help:
private static long countDuplicates(Collection<String> ids, UniquenessValidator validator) {
      return ids.stream().filter(validator::isUnique).count();
  }

or just use Arrays.asList("12.1", "1234.1", "111.1", "12.12")..filter(validator::isUnique).count(); ?


shouldn't we 'assert' after each partial test?

xml-occurrence-parsers/src/test/java/org/gbif/xml/occurrence/parser/parsing/validators/UniquenessValidatorTest.java
```diff
@@ -0,0 +1,89 @@
...
+      }
+
+      // add more values to test with high volume of data
+      Stopwatch watch = Stopwatch.createStarted();
+      for (int i = 0; i < n; i++) {
+        if (!validator.isUnique(UUID.randomUUID().toString())) {
+          duplicatesFound++;
+        }
+      }
+      LOG.info("time: " + watch.stop().elapsed(TimeUnit.MILLISECONDS));
```
fmendezh: LOG.info("time {}", watch.stop().elapsed(TimeUnit.MILLISECONDS));

xml-occurrence-parsers/src/main/java/org/gbif/xml/occurrence/parser/parsing/extendedrecord/ExtendedRecordConverter.java
asturcon: finally removed since it's only used in this parser. All tests passed.

PR #97, opened 2018-04-16
93 assemble pipelines

### LINKED ISSUE #93: Execute/assemble interpretation pipelines selectively based on an input parameter
Implement a mechanism to specify which interpretation sub-pipelines are executed, for example: ALL (Default), Taxonomy, Geospatial, Temporary, etc. It must possible to specify 1 or more sub-pipelines to be executed. This mechanism must handles the configuration settings required by each sub-pipeline.

asturcon:
closes #93


beam-common/src/main/java/org/gbif/pipelines/config/EsProcessingPipelineOptions.java
```diff
@@ -0,0 +1,90 @@
...
+import org.apache.beam.sdk.options.Default;
+import org.apache.beam.sdk.options.DefaultValueFactory;
+import org.apache.beam.sdk.options.Description;
+import org.apache.beam.sdk.options.PipelineOptions;
+
+@Experimental(Experimental.Kind.FILESYSTEM)
+public interface EsProcessingPipelineOptions extends DataProcessingPipelineOptions{
+
+  String[] DEFAULT_ES_HOSTS =
+    new String[] {"http://c4n1.gbif.org:9200", "http://c4n2.gbif.org:9200", "http://c4n3.gbif.org:9200",
```
fmendezh: we must remove these defaults

beam-common/src/main/java/org/gbif/pipelines/config/InterpretationType.java
```diff
@@ -0,0 +1,17 @@
...
+import com.google.common.collect.ImmutableList;
+
+/**
+ * Defines the types of the interpretations.
+ */
+public enum InterpretationType {
+
+  ALL, LOCATION, TEMPORAL, TAXONOMY, COMMON;
+
+  public static final List<InterpretationType> ALL_INTERPRETATIONS =
```
fmendezh: use public static final EnumSet<InterpretationType> ALL_INTERPRETATIONS =
    EnumSet.of(LOCATION, TEMPORAL, TAXONOMY, COMMON); 
instead, why we need ALL_INTERPRETATION if we have an ALL literal

beam-common/src/main/java/org/gbif/pipelines/config/EsProcessingPipelineOptions.java
asturcon: removed

beam-common/src/main/java/org/gbif/pipelines/config/InterpretationType.java
asturcon: changed

Issue #99, opened 2018-04-24
Compare our interpretations with the current ones (v1) and add the missing ones


asturcon:
So far, the missing fields identified are:
* DwcTerm.stateProvince in LocationInterpreter
* DcTerm.modified and DwcTerm.dateIdentified in TemporalInterpreter
* DcTerm.references
* Multimedia

This task includes to identify any other missing fields.

The logic of the interpretation in our pipeline has to be the same as the current one.

PR #102, opened 2018-05-04
99 missing interpretations

asturcon:
Added interpretations for:
- DwcTerm.stateProvince
- DcTerm.modified
- DwcTerm.dateIdentified (also added parsing of dateIdentified in taxonomy interpretation)
- DcTerm.references 
- Multimedia

### DIFF
--- a/core/src/main/java/org/gbif/pipelines/core/parsers/legacy/CountryMaps.java
+++ b/core/src/main/java/org/gbif/pipelines/core/parsers/legacy/CountryMaps.java
@@ -23,8 +23,6 @@ public class CountryMaps {
 
   private static final String CONFUSED_COUNTRY_FILE = "confused-country-pairs.txt";
 
-  private CountryMaps() {}
-
   /*
    Some countries are commonly mislabeled and are close to correct, so we want to accommodate them, e.g. Northern
    Ireland mislabeled as Ireland (should be GB). Add comma separated pairs of acceptable country swaps in the
@@ -35,6 +33,8 @@ public class CountryMaps {
   private static final Map<Country, Set<Country>> CONFUSED_COUNTRIES = Maps.newHashMap();
   // And this is the same, but without the issue — we aren't exactly following ISO, but we accept it.
   private static final Map<Country, Set<Country>> EQUIVALENT_COUNTRIES = Maps.newHashMap();
+
+  private CountryMaps() {}
 
   static {
     try (InputStream in = CountryMaps.class.getClassLoader().getResourceAsStream(CONFUSED_COUNTRY_FILE);
@@ -63,7 +63,7 @@ public class CountryMaps {
     Map<Country, Set<Country>> map = withIssue ? CONFUSED_COUNTRIES : EQUIVALENT_COUNTRIES;
 
     if (!map.containsKey(countryA)) {
-      map.put(countryA, Sets.<Country>newHashSet());
+      map.put(countryA, Sets.newHashSet());
     }
 
     Set<Country> confused = map.get(countryA);
--- a/core/src/main/java/org/gbif/pipelines/core/utils/ExtendedRecordCustomBuilder.java
+++ b/core/src/main/java/org/gbif/pipelines/core/utils/ExtendedRecordCustomBuilder.java
@@ -1,10 +1,17 @@
 package org.gbif.pipelines.core.utils;
 
+import org.gbif.api.vocabulary.Extension;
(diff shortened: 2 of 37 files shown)


asturcon:
closes #99

core/src/main/java/org/gbif/pipelines/core/parsers/legacy/CountryMaps.java
```diff
@@ -63,7 +63,7 @@ private static void addConfusedCountry(Country countryA, Country countryB, boole
     Map<Country, Set<Country>> map = withIssue ? CONFUSED_COUNTRIES : EQUIVALENT_COUNTRIES;
 
     if (!map.containsKey(countryA)) {
```
muttcg: A map has the method - putIfAbsent, you could use it instead of the check

core/src/main/java/org/gbif/pipelines/core/utils/ExtendedRecordCustomBuilder.java
```diff
@@ -129,41 +138,214 @@ public ExtendedRecordCustomBuilder infraspecificEpithet(String infraspecificEpit
...
+    return this;
+  }
+
   public ExtendedRecordCustomBuilder id(String id) {
     this.id = id;
     return this;
   }
 
+  public ExtendedRecordCustomBuilder addExtensionRecord(Extension extension, Map<String, String> record) {
+    if (extension != null) {
```
muttcg: Objects.nonNull(extension) ? :)

core/src/main/java/org/gbif/pipelines/core/utils/ExtendedRecordCustomBuilder.java
```diff
@@ -129,41 +138,214 @@ public ExtendedRecordCustomBuilder infraspecificEpithet(String infraspecificEpit
...
+
+    if (Objects.nonNull(extensions)) {
+      builder.setExtensions(extensions);
+    }
+
+    return builder.build();
+  }
+
+  private static void addToMap(Map<String, String> map, Term term, String value) {
     if (value != null) {
```
muttcg: Objects.nonNull(extension) ? :)

core/src/test/java/org/gbif/pipelines/core/parsers/multimedia/MultimediaParserTest.java
```diff
@@ -0,0 +1,265 @@
...
+        .derivedFrom("derivedFrom")
+        .audience("audience")
+        .rightsHolder("rightsHolder")
+        .creator("creator")
+        .format("image/jpeg")
+        .created("10/10/2000")
+        .build();
+
+    ExtendedRecord extendedRecord =
+      new ExtendedRecordCustomBuilder().id("123").addExtensionRecord(Extension.AUDUBON, audubonExtension).build();
```
muttcg: I think, would be better add method in ExtendedRecordCustomBuilder, like:
```
public static ExtendedRecordCustomBuilder create(){
    return new ExtendedRecordCustomBuilder();
}
```
It will be like - `ExtendedRecordCustomBuilder.create().id("123")`

core/src/test/java/org/gbif/pipelines/core/parsers/multimedia/MultimediaParserTest.java
```diff
@@ -0,0 +1,265 @@
...
+                 multimedia.getIdentifier().toString());
+  }
+
+  /**
+   * If the information about the same image (same URI) is given in the core AND the extension we should use the one
+   * from the extension (richer data).
+   */
+  @Test
+  public void testExtensionsPriority() {
+    Map<String, String> audubonExtension = new ExtendedRecordCustomBuilder.MultimediaExtensionBuilder().accessURI(
```
muttcg: For it looks a bit weird, I lost context why it is a map, not a MultimediaExtension class, maybe would be better create a class - MultimediaExtension class

asturcon: It is because in the ExtendedRecord the extensions are a map and every extension has a list of maps (as I understood it's because one record can have several extension rows associated to it):

```
java.util.Map<java.lang.String,java.util.List<java.util.Map<java.lang.String,java.lang.String>>> extensions;
```

and here I'm just creating a ExtendedRecord with extensions for testing. This map is like the core terms map.

PR #104, opened 2018-05-08
Checkstyle and xlint plugins enabled in pom

aalbatross:


aalbatross:
I have attached checkstyle and suppressions settings from apache beam project. Which could be changed on need.

The build currently is configured to fail in case of check-style violations.

Issue #114, opened 2018-06-06
Elasticsearch index creation, swapping and aliasing  mechanism


fmendezh:
This mechanism/component must provide services/methods to:
1. Create an index for each crawl attempt, this index should be created using settings recommended to improve indexing time (https://www.elastic.co/guide/en/elasticsearch/reference/master/tune-for-indexing-speed.html)
2. Index swapping: a newly created index replace and index being used, this can be done thru the ES Alias API using atomic updates, the new Index should use settings recommended for searching speed.

PR #126, opened 2018-06-12
ES indexing module

asturcon:
Pending to add mappings. Needs to be discussed


asturcon:
mappings added 
closes #114

Issue #158, opened 2019-05-14
Modify the HBase export to save a file per dataset


timrobertson100:
The current HBase export beam pipeline produces an extended record avro file for everything.
This should create a file per dataset using the FileIO dynamic destinations approach if possible.

PR #159, opened 2019-05-14
Creates a file per dataset for the HBase export (issue #158)

timrobertson100:
This is the first steps towards addressing #158 but still needs some work before closing the issue (work outstanding is to provide packaging to run on the cluster, save to HDFS, parameterise the sharding of the avro files and to allow a parameter to exclude datasets to avoid reprocessing e.g. eBird).


timrobertson100:
I think the second and third commits resolves the conflicts but it will need forced @fmendezh

PR #256, opened 2020-03-09
Fragmenter

muttcg:
PR contains many files, but you need to review only this module https://github.com/gbif/pipelines/tree/fragmenter/pipelines/ingest-fragments-uploader


muttcg:
Thanks, @timrobertson100, I applied your suggestions

PR #317, opened 2020-07-03
Allow HDFS location to be configurable

### LINKED ISSUE #294: Allow HDFS location to be configurable in FileSystemFactory
FileSystemFactory is currently hardcoded to the address hdfs://ha-nn

https://github.com/gbif/pipelines/blob/master/pipelines/ingest-gbif/src/main/java/org/gbif/pipelines/ingest/utils/FileSystemFactory.java#L16

This value is used to decide if a file system path is HDFS or not. Anything not starting with this path is deemed to local storage.

djtfmartin:
Allow HDFS location to be configurable in FileSystemFactory  - fixes #294


djtfmartin:
thanks @muttcg  - Ive updated this patch

PR #323, opened 2020-07-07
Remove hardcoded hdfs prefix

muttcg:
@djtfmartin your latest changes broke our ingestion, I fixed it and removed hardcoded hdfs prefix everywhere, the fix should work with your code without any changes, please review changes

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 sudo: false
 language: java
 jdk:
-  - oraclejdk8
+  - openjdk8
 branches:
   only:
   - master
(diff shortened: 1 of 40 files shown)


djtfmartin:
@muttcg can you share examples of `core-site.xml` and `hdfs-site.xml` ?

timrobertson100:
> @muttcg can you share examples of `core-site.xml` and `hdfs-site.xml` ?

I've done this in private comms

djtfmartin:
thanks @timrobertson100 

The supplied XML docs are big - i guess the better question is what are mandatory params in each ?
i.e. what was missing that broke things in your environment that other users should be aware of...

muttcg:
> thanks @timrobertson100
> 
> The supplied XML docs are big - i guess the better question is what are mandatory params in each ?
> i.e. what was missing that broke things in your environment that other users should be aware of...

hdfsSiteConfig returns some default FS when you do config.get("fs.default.name"), while value in xml doesn't exist
but in our case coreSiteConfig contains "fs.default.name", core-site.xml and hdfs-site.xml are generated by Cloudera

djtfmartin:
I think ive answered my own question - i can see the code is now checking for `fs.defaultFS` in both hdfs-site.xml and core-site.xml

PR #326, opened 2020-07-16
Merge of livingatlas pipelines

djtfmartin:
I've opened this PR now to progress the conversation on code formatting and standards. There is likely to be lots of conversation here.

The bulk of the development work includes:
1. Separate pipelines for interpretation, sampling, UUIDs and SOLR indexing
1. ALA taxonomy integration
1. ALA Collectory integration

Changes on this branch from recent work in the separate repository (https://github.com/atlasoflivingaustralia/la-pipelines) include:

1. Using fabric8 maven plugin to initialise docker instances for integration tests
2. Use of GBIF parent POM and rationalisation of dependencies to use inherited versions where possible.
3. Changes to tests to separate unit tests from integration tests.

The majority of the changes are solely in the `livingatlas` module, the following exceptions:
1. There are additions to AVRO models to support ALA - notably taxonomy.
1. A bug fix to DwcaIO class to allow for loading from a uncompressed DwcA with Beam's DirectRunner.

For more context, there is an open issue here that is used for tracking the work required to get the `livingatlas` module into shape for this merge:

https://github.com/AtlasOfLivingAustralia/la-pipelines/issues/93


livingatlas/migration/pom.xml
```diff
@@ -0,0 +1,187 @@
...
+                                <!-- append default configs -->
+                                <transformer implementation="org.apache.maven.plugins.shade.resource.AppendingTransformer">
+                                    <resource>reference.conf</resource>
+                                </transformer>
+                            </transformers>
+                            <relocations>
+                                <!-- Transient from core: conflicts with Beam on Spark -->
+                                <relocation>
+                                    <pattern>com.fasterxml.jackson</pattern>
+                                    <shadedPattern>jjjj213.com.fasterxml.jackson</shadedPattern>
```
timrobertson100, line 64: Suggest remove jjj

livingatlas/migration/pom.xml
```diff
@@ -0,0 +1,187 @@
...
+                        <groupId>com.google.errorprone</groupId>
+                        <artifactId>error_prone_core</artifactId>
+                        <version>${error_prone_core.version}</version>
+                    </dependency>
+                </dependencies>
+            </plugin>
+        </plugins>
+
+    </build>
+    <properties>
```
timrobertson100, line 130: Is it intentional that properties are overridden here?

livingatlas/pipelines/pom.xml
```diff
@@ -0,0 +1,890 @@
...
+              <unpack>true</unpack>
+              <outputDirectory>/tmp/pipelines-shp</outputDirectory>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
+
+      <!-- This configuration is required to override the version in the GBIF mother POM
+      which causes OOM issues due to its use of forking.
+      -->
```
timrobertson100, line 70: Suggest add "See https://github.com/gbif/pipelines/issues/327"

livingatlas/pipelines/pom.xml
```diff
@@ -0,0 +1,890 @@
...
+              <alias>ala-nameservice</alias>
+              <name>djtfmartin/ala-namematching-service:v20200214-4</name>
+              <external>
+                <type>compose</type>
+                <composeFile>ala-nameservice.yml</composeFile>
+              </external>
+              <run>
+                <wait>
+                  <http>
+                    <url>http://localhost:9179/api/search?q=Acacia</url>
```
timrobertson100, line 205: This looks suspicious - what if the port is in use by another test running on Jenkins? Should it be port scanning and assigning ports dynamically?

livingatlas/pipelines/pom.xml
```diff
@@ -0,0 +1,890 @@
...
+            </image>
+          </images>
+        </configuration>
+      </plugin>
+    </plugins>
+  </build>
+
+  <properties>
+    <!-- Common variables -->
+    <jdkLevel>1.8</jdkLevel>
```
timrobertson100, line 221: Is this different from a parent pom?

livingatlas/pipelines/pom.xml
```diff
@@ -0,0 +1,890 @@
...
+    <jcommander.version>1.78</jcommander.version>
+    <layers-store.version>2.0.3</layers-store.version>
+
+    <!-- Plugins -->
+    <io.fabric8.version>0.33.0</io.fabric8.version>
+    <org.jetbrains.annotations>13.0</org.jetbrains.annotations>
+
+  </properties>
+
+  <scm>
```
timrobertson100, line 250: This looks suspicious

livingatlas/pipelines/pom.xml
```diff
@@ -0,0 +1,890 @@
...
+    <jdkLevel>1.8</jdkLevel>
+    <encoding>UTF-8</encoding>
+    <surefireArgLine>-Xmx1024m</surefireArgLine>
+
+    <project.build.sourceEncoding>${encoding}</project.build.sourceEncoding>
+    <project.reporting.outputEncoding>${encoding}</project.reporting.outputEncoding>
+    <project.resources.sourceEncoding>${encoding}</project.resources.sourceEncoding>
+
+    <!-- Tools-->
+    <gbif-pipelines.version>2.5.4-SNAPSHOT</gbif-pipelines.version>
```
timrobertson100, line 230: This seems suspicious, as the module is part of the gbif-pipelines project

livingatlas/pipelines/pom.xml
```diff
@@ -0,0 +1,890 @@
...
+      <artifactId>lombok</artifactId>
+      <version>${lombok.version}</version>
+      <scope>provided</scope>
+    </dependency>
+
+    <!-- This project -->
+    <dependency>
+      <groupId>org.gbif.pipelines</groupId>
+      <artifactId>models</artifactId>
+      <version>${gbif-pipelines.version}</version>
```
timrobertson100, line 320: Versions can be removed here, as they're defined in parent poms

livingatlas/pipelines/pom.xml
```diff
@@ -0,0 +1,890 @@
...
+      <exclusions>
+        <exclusion>
+          <groupId>com.beust</groupId>
+          <artifactId>jcommander</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+
+    <!-- GBIF -->
+    <dependency>
```
timrobertson100, line 408: Versions should be removed here I think

livingatlas/pipelines/pom.xml
djtfmartin: The version of the module is set to `1.0-SNAPSHOT`. If this is not set, then maven assumes the gbif dependencies are `1.0-SNAPSHOT` as well.

livingatlas/migration/pom.xml
```diff
@@ -0,0 +1,187 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <groupId>au.org.ala</groupId>
+        <artifactId>livingatlas</artifactId>
+        <version>1.0-SNAPSHOT</version>
```
timrobertson100, line 8: Should be same as the project?

livingatlas/pipelines/pom.xml
djtfmartin: The version specified in GBIF parent POMs is the ${project.version}. The module version here is 1.0-SNAPSHOT.

livingatlas/pipelines/pom.xml
djtfmartin: No idea how to achieve that.

livingatlas/migration/pom.xml
djtfmartin: is it required to be ?

livingatlas/migration/pom.xml
djtfmartin: No. ill remove.

livingatlas/pipelines/pom.xml
timrobertson100: > The version of the module is set to 1.0-SNAPSHOT.

I added a comment that also seems wrong. This module is part of the 2.5.4 release of pipelines, and I think it'll be confusing to vary versions across modules and might break the current release tools. Are there examples of other projects doing this?

djtfmartin: I thought this was being done with a module called `ingest-hdfs-table` that i thought was within this project. Perhaps its now been moved ?

djtfmartin: ive switched to 2.5.4-SNAPSHOT

livingatlas/pipelines/pom.xml
djtfmartin: switched to 2.5.4-SNAPSHOT and removed versions

livingatlas/pipelines/pom.xml
djtfmartin: removed.

livingatlas/migration/pom.xml
djtfmartin: Removed

livingatlas/migration/pom.xml
djtfmartin: removed 3 of js

livingatlas/pipelines/pom.xml
djtfmartin: done

livingatlas/pipelines/pom.xml
djtfmartin: done

livingatlas/pipelines/pom.xml
timrobertson100: We can look. In jenkins you can do things like this, but the app needs to allow you provide a port:

Screenshot 2020-07-16 17 17 14

In code, we do things [like this](https://github.com/gbif/registry/blob/3b7b97c2f7dc369de7d465753f41b6447d8b0415/registry-cli/src/test/java/org/gbif/registry/cli/datasetindex/DatasetBatchIndexerIT.java#L68)

Can you please see if it's possible to pass in a port at startup? It's a nuisance doing a release and 20 minutes in you get a port collision that requires manual intervention to rollback. If not, we should open an issue for someone to investigate this.

djtfmartin: Please share a bit more of this example in the screenshot, including the actual POM in use (a snippet if the not the whole) and we'll try and copy this pattern.

djtfmartin: i should add that I don't currently have access to the environment to test this.

qifeng-bai: Can we run 'docker-compose down' to stop all running containers before we run tests?

MattBlissett: It's the registry project: https://github.com/gbif/registry/blob/master/pom.xml

I think `jetty.port` is probably used automatically by Jetty, I don't see any reference to it in our code. `System.getProperty`, anyway.

MattBlissett: The build command in Jenkins is now `mvn clean install verify dependency:analyze $SONAR_MAVEN_GOAL -Dhttp.port=$HTTP_PORT -Dsonar.host.url=$SONAR_HOST_URL -Dsonar.login=$SONAR_AUTH_TOKEN -U -P coverage`.

timrobertson100: Please bear in mind I only looked quickly.

The examples I listed were to show that Jetty can assign TCP PORT randomly per execution as a system variable. Here I think this relates to port mapping for docker containers, which are defined in the YAML? That may become much more convoluted (e.g. templating the prop files) unless the maven plugin lets us provide that from system variables.

If it is convoluted, I suggest we open an issue to "Parameterize the port in docker tests to allow concurrent executions" and then comment out their execution with a reference to the issue to let this progress.

Sound reasonable?

djtfmartin: Sorry - do you mean comment out the execution of the docker containers ?

timrobertson100: If it is the case that concurrent builds in jenkins all use the same ports in their tests, then yes (i.e. disable the tests until we solve that). If Jenkins doesn't run these, it can be ignored completely.

timrobertson100: Fair enough. It needs to be solved at some point. 
It was only if we cared to delay it to not block this merge.

djtfmartin: So to confirm, this needs to fixed before a merge is accepted ?

timrobertson100: https://github.com/gbif/pipelines/issues/337 I'll look into this, but others are welcome too

djtfmartin: this is resolved with the work on #337.

@qifeng-bai the maven build uses different ports for the docker containers to the ports specified in the docker-compose files described in the "Getting Started" section. This separation should help avoid the ports classh with maven builds from dev work in IDEs.

livingatlas/pipelines/pom.xml
djtfmartin: i presume this was the jackson version which was using the range. This has been changed.

PR #341, opened 2020-07-22
330 contributing guidelines

timrobertson100:
Based on private communications and group chat I have tried to compile the start of a contributing guide.
This is a combination of what Apache Beam uses and the existing material in this project.

Because this affects everyone, it needs consideration before applying.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,14 +2,14 @@
 
 The GBIF Pipelines development community welcomes contributions from anyone! Thank you.
 
-If you have questions please open [an issue](https://github.com/gbif/pipelines/issues/new) or reach out to the developers starting with informatics@gbif.org.
+If you have questions please open [an issue](https://github.com/gbif/pipelines/issues/new) or contact the developers at informatics@gbif.org.
 
-There are different ways you can contribute that heklp this project:
-- Log [issues](https://github.com/gbif/pipelines/issues) and help document a bug or specify new functionality 
+There are different ways you can contribute that help this project:
+- Log [issues](https://github.com/gbif/pipelines/issues) and help document a bug or specify new functionality
 - Improve documentation
-- Provide code submissions that address issues or brings new functionalities
+- Provide code submissions that address issues or bring new functionalities
 - Help test the functionalities offered in the project
-- Help improve this guide 
+- Help improve this guide
 
 ## Contributing code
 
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,119 @@
+# Contributing to GBIF Pipelines
+
+The GBIF Pipelines development community welcomes contributions from anyone! Thank you.
+
+If you have questions please open [an issue](https://github.com/gbif/pipelines/issues/new) or contact the developers at informatics@gbif.org.
```
djtfmartin, line 5: As Matt has suggested, perhaps we need a mailing list for GBIF and ALA devs.

CONTRIBUTING.md
```diff
@@ -0,0 +1,119 @@
...
+#### Command line
+
+This project uses [Apache Maven](https://maven.apache.org/run.html) to build. The following should be enough to checkout and build the project.
+
+```
+git clone git@github.com:gbif/pipelines.git
+cd pipelines
+mvn clean package
+```
+To skip integration tests, run `mvn clean package -DskipITs`.
```
muttcg, line 43: "mvn clean package" won't run IT, only UT, "verify" runs IT

CONTRIBUTING.md
```diff
@@ -0,0 +1,119 @@
...
+
+#### IDE Setup
+
+We recommend the following toolset, and are able to answer questions on this configuration:
+
+- Use [IntelliJ IDEA Community](https://www.jetbrains.com/idea/download/) (or better),
+- Use the [Google Java Format](https://plugins.jetbrains.com/plugin/8527-google-java-format). Please do not reformat existing code, only changes and new code.
+- The project uses [Project Lombok](https://projectlombok.org/). Please install the [Lombok plugin for IntelliJ IDEA](https://plugins.jetbrains.com/plugin/6317-lombok-plugin).
+- Because the project uses [Error Prone](https://code.google.com/p/error-prone) you may have issues during the build process from IDEA.  To avoid these issues please install the [Error Prone compiler integration plugin](https://plugins.jetbrains.com/plugin/7349-error-prone-compiler-integration) and build the project using the [`error-prone java compiler`](https://code.google.com/p/error-prone) to catch common Java mistakes at compile-time. To use the compiler, go to _File_ → _Settings_ → _Compiler_ → _Java Compiler_ and select `Javac with error-prone` in the `Use compiler` box.
```
djtfmartin, line 56: Theres ongoing issues with error prone and Intellij (see slack discussion with Vicente - i have  had the same issues)

CONTRIBUTING.md
```diff
@@ -0,0 +1,119 @@
...
+Below is a tutorial for contributing code to Pipelines, covering our tools and typical process in detail.
+
+### Prerequisites
+
+To contribute code, you need
+
+- a GitHub account
+- a Linux, MacOS, or Microsoft Windows development environment with Java JDK 8 installed
+- Maven (version 3.6+)
+
```
muttcg, line 25: Maybe specify Java 8 version? I tried to use jdk 11 and target 8, and it doesn't work

CONTRIBUTING.md
timrobertson100: Thanks. I suggest we also get mavenized `errorprone` options working too for the reason it is more robust than relying on IDE use. I haven't explored that yet

CONTRIBUTING.md
timrobertson100: My thoughts were not to at this point. If we find it is used and there is a need to, we can create one at short notice and change this guide but I expect we won't. Initially, I expect we'd be able to guide people towards creating issues and keeping discussion there and joining slack instead.

Happy to create one now if others think it is necessary at this stage.

CONTRIBUTING.md
```diff
@@ -0,0 +1,119 @@
...
+    2. Examples of bad commit messages:
+          1. ` `
+          1. `Various fixes`
+          1. `#123`
+5. Check your code compiles, and all project unit tests pass (please be kind to reviewers)
+6. Explore the `errorprone` warnings raised at compilation time. Please address issues you see as best you can.
+7. Ensure that the code is spotless (`mvn spotless:check` and fixup styling issues using `mvn spotless:apply`)
+8. Verify that the PR only changes the code necessary to address the issue (other fixes should be in separate PRs)
+6. Prefer to create a pull request and have it reviewed for larger submissions. Committers are free to push *smaller* changes directly.
+
```
muttcg, line 92: Should we add "Delete the branch after PR"? I am not sure that a user will be able to do that without some permissions

MattBlissett: Perhaps the person with permission to merge should do that.

CONTRIBUTING.md
djtfmartin: Maybe a mailing list will give a better impression of building a community. It would also allow Living Atlas coordinators to follow conversations.

timrobertson100: pipelines@lists.gbif.org ?

Issue #324, opened 2020-07-14
Add GADM interpretation to records with coordinates


MattBlissett:
To enable search and analysis by administrative region, add GADM at levels 0, 1, 2 and 3 to occurrences.

This should be an additional field, and not change or use `dwc:stateProvince` etc.

Depends on https://github.com/gbif/geocode/issues/6

timrobertson100:
Please consider that we anticipate GADM will be used a lot (e.g. create a choropleth map with an aggregation code, metrics for country pages aggregated by county) which may influence the ES structure necessary.

MattBlissett:
GADM will give us this sort of data:

```
 type  │     id      │      source      │ title  │ isocountrycode2digit 
───────┼─────────────┼──────────────────┼────────┼──────────────────────
 GADM0 │ JPN         │ http://gadm.org/ │ Japan  │ JP
 GADM1 │ JPN.26_1    │ http://gadm.org/ │ Nagano │ JP
 GADM2 │ JPN.26.40_1 │ http://gadm.org/ │ Nagawa │ JP

```

The ids are clearly structured. @fmendezh, is there something ES can do with this, or should it be four fields (some countries will have `GADM3`)?

----

Do we want to index the title, or just the id?

timrobertson100:
I think we need code and title since we'll want to include this in the occurrence JSON response. 

I could imagine all these being useful:

```
"GADM": {
  isoCountryCode:JP,      // to help spot possible data errors
  level0: {
    "code": "JPN",
    "title": "Japan"
  },
  level1: {
    "code": "JPN.26_1",
    "title": "Nagano"
  },
  level2: {
    "code": "JPN.26.40_1",
    "title": "Nagawa"
  }
}
```

We will definitely want to be able to search and aggregate counts by code, and I don't know if flattening data for ES will help with that (e.g. holding `gadm0Code` and `gadm0Title`).

MattBlissett:
How important is the `isoCountryCode`?

GADM uses three letter codes (which we could map to two letters using `Country`), but it also includes an additional 7 custom codes:

```
 XAD   │ Akrotiri and Dhekelia
 XCA   │ Caspian Sea          
 XCL   │ Clipperton Island    
 XKO   │ Kosovo                 (NB we already have XKX from other sources)
 XNC   │ Northern Cyprus      
 XPI   │ Paracel Islands      
 XSP   │ Spratly Islands      
```

[...]

PR #344, opened 2020-07-22
324 Add GADM

MattBlissett:
This adds interpretation through the existing geocode KVS, and handles the HDFS table (NB needs branch version of occurrence).

@fmendezh / @muttcg, advice requested.

We have eight new fields:
- level0gid, level0name
- level1gid, level1name
- level2gid, level2name
- level3gid, level3name

We'll get this sort of data:

```
 type  │    gid      │      source      │ name  
───────┼─────────────┼──────────────────┼────────
 GADM0 │ JPN         │ http://gadm.org/ │ Japan   
 GADM1 │ JPN.26_1    │ http://gadm.org/ │ Nagano  
 GADM2 │ JPN.26.40_1 │ http://gadm.org/ │ Nagawa 
```

 As Tim mentioned at https://github.com/gbif/pipelines/issues/324#issuecomment-658253127

> Please consider that we anticipate GADM will be used a lot (e.g. create a choropleth map with an aggregation code, metrics for country pages aggregated by county) which may influence the ES structure necessary.

Can you think about this? I don't know enough about ES to decide on the appropriate structure.

It's also just occurred to me that a user might search "長野県" or "長和町". Do we need this too?


sdks/parsers/src/main/java/org/gbif/pipelines/parsers/parsers/location/parser/LocationMatcher.java
```diff
@@ -131,6 +142,24 @@ public LocationMatcher additionalTransform(UnaryOperator<LatLng> transformation)
     return Optional.empty();
   }
 
+  private Optional<GadmFeatures> getGadmFromCoordinates(LatLng latLng) {
+    if (latLng.isValid()) {
+      GeocodeResponse geocodeResponse = null;
+      try {
+        geocodeResponse = geocodeKvStore.get(latLng);
+      } catch (NoSuchElementException | NullPointerException ex) {
```
fmendezh, line 150: catching those exceptions here looks like there's bug somewhere else

sdks/parsers/src/main/java/org/gbif/pipelines/parsers/parsers/location/parser/LocationParser.java
```diff
@@ -99,6 +100,22 @@
...
+    Objects.requireNonNull(lr, "LocationRecord is required");
+    Objects.requireNonNull(kvStore, "GeocodeService kvStore is required");
+
+    // Take parsed values
+    LatLng location = new LatLng(lr.getDecimalLatitude(), lr.getDecimalLongitude());
+
+    // Use these to retrieve the GADM areas.
+    Optional<GadmFeatures> match =
+        LocationMatcher.create(location, null, kvStore)
+            .applyGadm();
```
muttcg, line 113: Could be replaced to 
```
return LocationMatcher.create(location, null, kvStore)
            .applyGadm();
```
or added debug logger

pipelines/ingest-transforms/src/test/java/org/gbif/pipelines/transforms/core/LocationTransformTest.java
```diff
@@ -53,9 +54,59 @@ public void processElement(ProcessContext context) {
   public final transient TestPipeline p = TestPipeline.create();
 
   private static GeocodeResponse toGeocodeResponse(Country country) {
-    Location location = new Location();
-    location.setIsoCountryCode2Digit(country.getIso2LetterCode());
-    return new GeocodeResponse(Collections.singletonList(location));
+    List<Location> locations = new ArrayList();
```
muttcg, line 57: Raw usage
`new ArrayList<>();`

sdks/parsers/src/main/java/org/gbif/pipelines/parsers/parsers/location/cache/GeocodeBitmapCache.java
```diff
@@ -24,8 +23,7 @@
   // World map image lookup
   private final BufferedImage img;
   private static final int BORDER = 0x000000;
-  private static final int EEZ = 0x888888;
-  private static final int INTERNATIONAL_WATER = 0xFFFFFF;
+  private final static int NOTHING = 0xFFFFFF;
```
muttcg, line 26: Java specs recommend "static final" order

sdks/parsers/src/main/java/org/gbif/pipelines/parsers/parsers/location/parser/GadmFeatures.java
```diff
@@ -0,0 +1,65 @@
...
+            level2Gid = l.getId();
+            level2Name = l.getName();
+          }
+          return;
+        case "GADM3":
+          if (level3Gid == null) {
+            level3Gid = l.getId();
+            level3Name = l.getName();
+          }
+          return;
```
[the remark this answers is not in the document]
MattBlissett: I won't remove 45, even here it's not great to use a fall-through.  [Google Code style](https://google.github.io/styleguide/javaguide.html#s4.8.4-switch) requires the `default:`, the `return` in it doesn't really matter.

I left it as there are other values coming through, like `EEZ` and `Political`.

(When compiled, I doubt any of these variations make any difference.)

sdks/parsers/src/main/java/org/gbif/pipelines/parsers/parsers/location/parser/LocationMatcher.java
MattBlissett: It's not required. I'd copied the call from the method above (the original geocoding) without checking.

Removing here and above.

pipelines/ingest-gbif/src/main/resources/elasticsearch/es-occurrence-schema.json
```diff
@@ -118,6 +118,20 @@
...
+        "type": "object",
+        "properties": {
+          "gadmLevel0Gid": {"type": "keyword"},
+          "gadmLevel1Gid": {"type": "keyword"},
+          "gadmLevel2Gid": {"type": "keyword"},
+          "gadmLevel3Gid": {"type": "keyword"},
+          "gadmLevel0Name": {"type": "text"},
+          "gadmLevel1Name": {"type": "text"},
+          "gadmLevel2Name": {"type": "text"},
+          "gadmLevel3Name": {"type": "text"}
```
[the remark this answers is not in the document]
MattBlissett: Should it be `gadm.level0Gid` then?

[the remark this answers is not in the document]
muttcg: Yes, I think it should be:
```
gadm
- level0Gid
- level0Name
- ..ect
```

Issue #96, opened 2018-04-10
Remove collection lineages and issues from current interpretation classes


fmendezh:
Since this milestone aims to re-use current interpreters and parsers the propose way of collecting lineages and issues is not required to produce a new datastore to enable GBIF API v1

PR #345, opened 2020-07-23
Fixes for issues #96 and #105 

djtfmartin:
Fix for AtlasOfLivingAustralia/la-pipelines#96 - making the initialisation of the centre points objects and vocab more thread safe by initialising in the transform.

Temporary fix for AtlasOfLivingAustralia/la-pipelines#105 while we investigate the resource cleanup issues.


livingatlas/pipelines/src/main/java/au/org/ala/pipelines/interpreters/ALALocationInterpreter.java
```diff
@@ -94,124 +91,54 @@
...
-                    lr.getDecimalLongitude(),
-                    lr.getStateProvince());
-              }
+          if (stateProvinceCentrePoints.coordinatesMatchCentre(
+              lr.getStateProvince(), lr.getDecimalLatitude(), lr.getDecimalLongitude())) {
+            addIssue(lr, ALAOccurrenceIssue.COORDINATES_CENTRE_OF_STATEPROVINCE.name());
+          } else {
+            if (log.isTraceEnabled()) {
+              log.trace(
+                  "{},{} is not the centre of {}!",
```
timrobertson100, line 137: Has this got a typo? Should it read "...is [at] the centre of {}" ?

djtfmartin: it has yes - good catch

djtfmartin: i.e. no issue needs to be raise....

timrobertson100: Indeed. I am blind

livingatlas/pipelines/src/main/java/au/org/ala/pipelines/transforms/LocationTransform.java
```diff
@@ -79,6 +84,17 @@ public void setup() {
...
     }
+
+    try {
+      countryCentrePoints = CountryCentrePoints.getInstance(alaConfig.getLocationInfoConfig());
+      stateProvinceCentrePoints =
+          StateProvinceCentrePoints.getInstance(alaConfig.getLocationInfoConfig());
+      stateProvinceVocab =
+          StateProvince.getInstance(alaConfig.getLocationInfoConfig().getStateProvinceNamesFile());
+    } catch (Exception e) {
+      log.error(e.getMessage(), e);
```
qifeng-bai, line 95: Shall we provide more hints to help users correct configuration file?

djtfmartin: @qifeng-bai Im wondering if we need to do this now, as by default the vocab files that are shipped in the JAR will be used unless the user has configured some externally via YAML.

I think when we originally added this help, the vocab files would only be loaded externally so we needed the error messages.

qifeng-bai: I am OK without hints at this moment.

Issue #253, opened 2020-02-12
Improve recordedBy tokenization


timrobertson100:
Submitting by proxy (CC @emhaston @qgroom)

Searching by `recordedBy` causes frustration. It currently seems to operate as a strict prefix-only filter so a search for "Robertson" does not catch "Tim Robertson"

If we can tokenize that so that a search for `Robertson` could catch-all of `Tim Robertson`, `John Robertson` etc and `Robertson Tim` would boost the `Robertson, Tim`, `Tim Robertson`, `Timothy Robertson` in the results it would be greatly appreciated.

MattBlissett:
Searching for `T Robertson` could also bring back `Tim Robertson`, and `T X Robertson` might also be a useful partial match.

PR #347, opened 2020-07-24
Having these fields as text will allow to perform partial against them

fmendezh:
Changing these fields to be text will allow performing partial search matches against their values.
Be aware that this will require to change all mappings of those fields to be fieldName.keyword to preserve the current behavior (see https://github.com/gbif/occurrence/blob/master/occurrence-search/src/main/java/org/gbif/occurrence/search/es/OccurrenceEsField.java#L64)

https://github.com/gbif/pipelines/issues/253
https://github.com/gbif/occurrence/issues/39


MattBlissett:
Is there a way to fix https://github.com/gbif/portal-feedback/issues/2887 too?

(NB merge is currently to master. "Edit" looks like it can change that.)

fmendezh:
> Is there a way to fix [gbif/portal-feedback#2887](https://github.com/gbif/portal-feedback/issues/2887) too?
> 
> (NB merge is currently to master. "Edit" looks like it can change that.)
@MattBlissett  We can remove the lowercase_normalizer but we must be aware that all queries and facets will become case sensitive::

1.  A query like recordedBy=Matt it will be different to recorededBy=MATT
2.  Facets response can have now:  Matt: 100, and MATT:300  (which I think is what Niki wants)

muttcg:
Are you sure about occurrenceId, catalogNumber, and other identifiers?

timrobertson100:
It is common that people have e.g. "123 Robertson" for a field number, so it would be useful to have this tokenised search on occurrenceId, recordNumber, fieldNumber, catalogNumber and otherCatalogNumbers (which is often delimited in other ways=

muttcg:
> It is common that people have e.g. "123 Robertson" for a field number, so it would be useful to have this tokenised search on occurrenceId, recordNumber, fieldNumber, catalogNumber and otherCatalogNumbers (which is often delimited in other ways=

I see, thanks

@fmendezh Can we merge it into dev branch, instead of master? We will have to recreate the index with these changes, at the same time new features in dev require full reindexing

MattBlissett:
Changing the base branch on Github seems to have worked, it's now ready to merge into gbif-dev.

Issue #352, opened 2020-07-27
Remove or ignore unused Elasticsearch fields


fmendezh:
@MortenHofft suggested that many fields, especially in the gbifClassification nested element, are not being used by any search-related functionality and are also not used to build a response, some of those fields can be either removed or at least not-indexed to avoid use unnecessary resources.

Some example of unused fields:
- gbifClassification.diagnostic.note
- gbifClassification.usageParsedName[basionymAuthorship,candidatus, combinationAuthorship, incomplete, indetermined, uninomial]
- etc.

PR #353, opened 2020-07-27
Adding explicit mapping to the gbifClassification object to avoid ind…

fmendezh:
Adding explicit mappings to gbifClassification (taxonomy) to avoid indexing unnecessary data.

All fields in https://github.com/gbif/occurrence/blob/master/occurrence-search/src/main/java/org/gbif/occurrence/search/es/OccurrenceEsField.java must be kept and explicitly mapped.

Worth checking this PR against the JSON Converter: https://github.com/gbif/pipelines/blob/gbif-dev/sdks/core/src/main/java/org/gbif/pipelines/core/converters/GbifJsonConverter.java#L419 

https://github.com/gbif/pipelines/issues/352


asturcon:
I was wondering if we could also remove the fields that we're currently not returning in the occurrence search [response](https://github.com/gbif/occurrence/blob/19ca54254eb5afd4d14318e2e5cced4a17862736/occurrence-search/src/main/java/org/gbif/occurrence/search/es/EsResponseParser.java#L411) since they are not gonna be indexed now either - it's mostly fields in the `usageParsedName` node.

PR #363, opened 2020-08-10
112 gbif km

qifeng-bai:
Using gbif-parser instead of its own parser

### DIFF
--- a/livingatlas/pipelines/src/main/java/au/org/ala/pipelines/parser/UncertaintyParser.java
+++ b/livingatlas/pipelines/src/main/java/au/org/ala/pipelines/parser/UncertaintyParser.java
@@ -4,9 +4,9 @@ import java.util.UnknownFormatConversionException;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import lombok.extern.slf4j.Slf4j;
-import org.elasticsearch.common.Strings;
 import org.gbif.common.parsers.core.ParseResult;
 import org.gbif.common.parsers.geospatial.MeterRangeParser;
+import com.google.common.base.Strings;
 
 /**
  * This is a port of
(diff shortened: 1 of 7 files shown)


livingatlas/pipelines/src/main/java/au/org/ala/pipelines/parser/UncertaintyParser.java
```diff
@@ -4,13 +4,16 @@
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import lombok.extern.slf4j.Slf4j;
+import org.elasticsearch.common.Strings;
```
djtfmartin, line 7: this is a strange import. i dont think this code should have elasticsearch dependencies.

livingatlas/pipelines/src/main/java/au/org/ala/pipelines/parser/UncertaintyParser.java
```diff
@@ -4,13 +4,16 @@
...
+import org.gbif.common.parsers.core.ParseResult;
+import org.gbif.common.parsers.geospatial.MeterRangeParser;
 
 /**
  * This is a port of
  * https://github.com/AtlasOfLivingAustralia/biocache-store/blob/master/src/main/scala/au/org/ala/biocache/parser/DistanceRangeParser.scala
  */
 @Slf4j
-public class DistanceRangeParser {
+public class UncertaintyParser extends MeterRangeParser {
```
djtfmartin, line 16: Im not sure i understand the name change. I dont think theres anything inherent about the values that are specific to coordinateUncertaintyInMeters. I suggest we change back to DistanceRangeParser as this also helps with traceability with biocache-store code..

qifeng-bai: There are some specific ways, like range 1km-10km, Uncertainty always uses the second number. Since GBIF parser already handle UOM conversion, ours is more like Uncertainty specific implementation. Having said that, I am OK with both names.

livingatlas/pipelines/src/main/java/au/org/ala/pipelines/parser/UncertaintyParser.java
qifeng-bai: For checking if string is empty or null. Changed to google.common now

PR #365, opened 2020-08-15
121 temp working dir dwca

djtfmartin:
Fix for https://github.com/AtlasOfLivingAustralia/la-pipelines/issues/121 - allowing working directory for GNU sort to be specified
Removal of use of /efs-mount-point for config files - relying on locally deployed config.


livingatlas/scripts/la-pipelines
```diff
@@ -313,7 +320,7 @@ function interpret () {
             $(toArg $PRE driver-memory) \
             --class $CLASS \
             --master $SPARK_MASTER \
-            --driver-java-options "-Dlog4j.configuration=file:/efs-mount-point/log4j.properties" \
+            --driver-java-options "-Dlog4j.configuration=file:$CONFIG_DIR/log4j.properties" \
```
vjrj, line 323: I think we should use `$logConfig` better, but, now I remember that you commented me that you have issues with colorized logs running in Spark caused by a Class not found exception.

Some related discussion:
https://stackoverflow.com/a/52740534
https://stackoverflow.com/a/47734870

So me merge this like this, and also we can open an additional issue and try to fix it using `$logConfig`.

livingatlas/configs/la-pipelines.yaml
```diff
@@ -6,15 +6,19 @@ run:
...
     # jar: we get the jar from our dev or production environment
     sparkTmp: /data/spark-tmp
+    dwcaTmp: /data/dwca-tmp
+    dwcaImportDir: /data/biocache-load
     sparkMaster: ""
   spark-cluster:
     jar: /efs-mount-point/pipelines.jar
     sparkTmp: /data/spark-tmp
-    sparkMaster: spark://aws-spark-quoll-1.ala:7077
+    sparkMaster: spark://localhost:7077
```
vjrj, line 22: `dwcaImportDir` is empty in `spark-cluster` so I get this error:

```
ls: cannot access '/dr*/dr*.zip': No such file or directory
```
running `./la-pipelines do-all all --dry-run --cluster`

livingatlas/scripts/la-pipelines
```diff
@@ -252,20 +256,22 @@ function dwca-avro () {
...
 
     if [[ ! -f  $dwca_dir ]]
     then
-        dwca_dir="/data/biocache-load/$dr.zip"
+        dwca_dir="$DWCA_IMPORT/$dr.zip"
         log.error "$dwca_dir does not exists on your filesystem."
         exit 1
     fi
 
+    export TMPDIR=$DWCA_TMP
```
vjrj, line 273: I don't find the use of `TMPDIR`. Is this necessary?

vjrj (changes requested):

livingatlas/scripts/la-pipelines
djtfmartin: Thanks @vjrj it is in use here:

https://github.com/gbif/gbif-common/blob/master/src/main/java/org/gbif/utils/file/FileUtils.java#L827

The tmp directory is a issue for large DwCA files as the default for sorting is to use `/tmp` which causes issues for large files.

livingatlas/configs/la-pipelines.yaml
djtfmartin: thanks @vjrj ill fix this

livingatlas/scripts/la-pipelines
vjrj: Ah, ok. This is why I did find the use with a `grep -r`.

Can you add a minor comment to take into account in the future?

djtfmartin: lve added a comment  - thank you

livingatlas/configs/la-pipelines.yaml
djtfmartin: Ive added a default dwcaImportDir and tested. all ok now.

livingatlas/scripts/la-pipelines
djtfmartin: Ive switched to using $logConfig and this is working fine.

livingatlas/scripts/la-pipelines
```diff
@@ -26,7 +26,12 @@ set -e # Stop on any failure
 if [[ $SCRIPTPATH == "/usr/bin" ]] ; then PROD=true ; else PROD=false ; fi
 
 if [[ $PROD = false ]]; then
-    VER=$(grep -oPm1 "(?<=<version>)[^<]+" "../pom.xml")
+  # grep with -P option doesnt work on Mac OSX - perl regex arent supported,
+  if [[ "$OSTYPE" == "darwin"*  ]]; then
+    VER=$(grep version ../pom.xml | grep -v '<?xml'| head -1 | sed 's/[[:space:]]//g' | sed -E 's/<[/]{0,1}version>//g')
```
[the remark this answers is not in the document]
vjrj: The current version of `la-pipelines` fails for me in ubuntu because of this (`VER` is empty). This works for me:
```
diff --git a/livingatlas/scripts/la-pipelines b/livingatlas/scripts/la-pipelines
index 9146b3e9..f5c78fa3 100755
--- a/livingatlas/scripts/la-pipelines
+++ b/livingatlas/scripts/la-pipelines
@@ -26,12 +26,7 @@ set -e # Stop on any failure
 if [[ $SCRIPTPATH == "/usr/bin" ]] ; then PROD=true ; else PROD=false ; fi
 
 if [[ $PROD = false ]]; then
-  # grep with -P option doesnt work on Mac OSX - perl regex arent supported,
-  if [[ "$OSTYPE" == "darwin"*  ]]; then
     VER=$(grep version ../pom.xml | grep -v '<?xml'| head -1 | sed 's/[[:space:]]//g' | sed -E 's/<[/]{0,1}version>//g')
-  else
-    VER=$(grep -om1 "(?<=<version>)[^<]+" "../pom.xml")
-  fi
 else
     VER=$(dpkg-query --show -f '${Version}' la-pipelines)
 fi
```

but if for some reason you want to maintain the `if darwin` , just add the `-P` again:
```
diff --git a/livingatlas/scripts/la-pipelines b/livingatlas/scripts/la-pipelines
index 9146b3e9..2412f5b9 100755
--- a/livingatlas/scripts/la-pipelines
+++ b/livingatlas/scripts/la-pipelines
@@ -30,7 +30,7 @@ if [[ $PROD = false ]]; then
   if [[ "$OSTYPE" == "darwin"*  ]]; then
     VER=$(grep version ../pom.xml | grep -v '<?xml'| head -1 | sed 's/[[:space:]]//g' | sed -E 's/<[/]{0,1}version>//g')
   else
-    VER=$(grep -om1 "(?<=<version>)[^<]+" "../pom.xml")
+    VER=$(grep -oPm1 "(?<=<version>)[^<]+" "../pom.xml")
   fi
 else
     VER=$(dpkg-query --show -f '${Version}' la-pipelines)
```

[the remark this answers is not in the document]
djtfmartin: thanks @vjrj  - i removed the check. agree, much nicer without it !

vjrj (approved):

djtfmartin:
thanks @vjrj

PR #367, opened 2020-08-24
Fix multi dr args issue

vjrj:
A fix for a bug detected by @djtfmartin when using multi drs arguments.


djtfmartin (approved):
LTGM - thanks @vjrj

PR #377, opened 2020-09-04
Move the Antarctic check before the geocode lookup.

MattBlissett:
Geocode lookup now returns seas/oceans (e.g. "Southern Ocean"), but the previous behaviour relied on an empty response.

This will fix the many reported mismatches for occurrences in the Southern Ocean, and the lack of marine occurrences on the map on https://www.gbif.org/country/AQ/summary


muttcg (approved):
LGTM

PR #383, opened 2020-09-08
#380 changed tests that were giving problems with the notIssues array…

### LINKED ISSUE #380: Change GbifJsonConverterTest to use JsonObject instead of String
This is necessary to prevent tests from failing because of random order of some fields, like the `notIssues` arrays.

marcos-lg:
I only changed the tests that were giving problems with the `notIssues` array order

closes #380


muttcg (approved):
Thanks for fixing this, Marcos

PR #385, opened 2020-09-10
#369 grscicoll integration in pipelines

### LINKED ISSUE #369: Integrate the collections lookup service to link occurrences with GrSciColl data
A [collections lookup service](https://github.com/gbif/registry/issues/204) was recently implemented to help us matching occurrence data with GrSciColl institutions and collections. 

Now we need to integrate it into pipelines. The integration can be done in the following way:

- The call to the WS will be integrated as a KV-Store. Still to be determined if we'll use HBase cache or any other cache.
- The interpreted data will be stored in a separate avro file so in the future we can reinterpret only this data.
- We index to ES the keys, codes and names of the institutions and collections matched so we can perform searches on these fields. @MortenHofft do you have any other field in mind?

marcos-lg:
Summary of changes:
- Added call to grscicoll lookup service thru kv-store. The response is stored in a new avro file (`grscicoll-record.avsc`) and a new interpretation type (GRSCICOLL).
- Added to ES index `institutionKey` and `collectionKey` under a `grscicoll` node. There are already 2 fields `institutionKey` and `collectionKey` in the root but they should be deleted (right now they are always empty).
- Removed `institutionKey` and `collectionKey` from HDFS view since we don't want these fields in downloads.
- Removed all classes related to  `TaggedValues` to avoid confusions.

This needs these changes https://github.com/gbif/gbif-api/pull/67 to compile and the latest snapshot of the `kv-store`.

closes #369


timrobertson100:
DO NOT MERGE THIS

@marcos-lg asked me to block this merge as there is still work pending and he's boarding a plane

sdks/models/src/main/avro/core/grscicoll-record.avsc
```diff
@@ -0,0 +1,119 @@
...
+    "namespace" : "org.gbif.pipelines.io.avro.grscicoll",
+    "type" : "enum",
+    "symbols" : [
+       "CODE_MATCH", "IDENTIFIER_MATCH", "ALTERNATIVE_CODE_MATCH", "NAME_MATCH", "PROBABLY_ON_LOAN", "INST_COLL_MISMATCH",
+       "COUNTRY_MATCH", "INSTITUTION_TAG", "COLLECTION_TAG", "COLLECTION_TO_INSTITUTION_TAG", "INSTITUTION_TO_COLLECTION_TAG"
+    ]
+  },
+  {
+    "name" : "IdentifierType",
+    "namespace" : "org.gbif.pipelines.io.avro.grscicoll",
```
muttcg, line 27: I think we should try to avoid ENUMs in avro, we will have 2 places in total for ENUMs -> gbif.api and avro, I see 2 possible issues:
1) it is highly possible to get an error when we add/change an ENUM in gbif-api library and forgot about ENUMs in avro, and error we can find in runtime only
2) What if we have avro files with old version of ENUMs and we edited ENUMs

All in all, I think we should store only ENUM's name and parse them in code

sdks/core/src/main/java/org/gbif/pipelines/core/parsers/grscicoll/GrscicollRecordConverter.java
```diff
@@ -0,0 +1,140 @@
...
+      builder.setIdentifiers(identifiers);
+    }
+
+    return builder.build();
+  }
+
+  private static List<Reason> convertReasons(
+      Set<org.gbif.api.model.collections.lookup.Match.Reason> reasons) {
+    if (reasons == null || reasons.isEmpty()) {
+      return null;
```
muttcg, line 97: For collection better to return Collections.emptyList()

sdks/core/src/main/java/org/gbif/pipelines/core/parsers/grscicoll/GrscicollRecordConverter.java
```diff
@@ -0,0 +1,140 @@
+package org.gbif.pipelines.core.parsers.grscicoll;
```
muttcg, line 1: The class must be in converters package:
sdks/core/src/main/java/org/gbif/pipelines/core/converters/GrscicollRecordConverter.java

gbif/pipelines/ingest-gbif-java/src/main/java/org/gbif/pipelines/ingest/java/pipelines/VerbatimToInterpretedPipeline.java
```diff
@@ -192,6 +194,12 @@ public static void run(InterpretationPipelineOptions options, ExecutorService ex
             .create()
             .counterFn(incMetricFn);
 
+    GrscicollTransform grscicollTransform =
+        GrscicollTransform.builder()
+            .kvStoreSupplier(GrscicollLookupKvStoreFactory.getInstanceSupplier(config))
+            .create()
+            .counterFn(incMetricFn);
```
muttcg, line 201: To make it work you must add metric to org.gbif.pipelines.ingest.java.metrics.IngestMetricsBuilder

muttcg (commented):
Please add more test cases for new classes like GrscicollRecordConverter and others where it is possible

sdks/core/src/main/java/org/gbif/pipelines/core/interpreters/core/GrscicollInterpreter.java
```diff
@@ -0,0 +1,139 @@
...
+import static org.gbif.pipelines.core.utils.ModelUtils.addIssue;
+import static org.gbif.pipelines.core.utils.ModelUtils.checkNullOrEmpty;
+import static org.gbif.pipelines.core.utils.ModelUtils.extractNullAwareValue;
+
+import com.google.common.annotations.VisibleForTesting;
+import java.util.function.BiConsumer;
+import lombok.AccessLevel;
+import lombok.NoArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.elasticsearch.common.Strings;
```
muttcg, line 12: Is it ok that is elasticsearch version?

muttcg (approved):
Looks good, Marcos, thanks

sdks/core/src/main/java/org/gbif/pipelines/core/interpreters/core/GrscicollInterpreter.java
marcos-lg: no, that was wrong. I just changed it

muttcg (approved):

PR #390, opened 2020-09-16
Index GADM GIDs into an array, for searching at a mixture of levels.

MattBlissett:
See https://github.com/gbif/gbif-api/issues/65

### DIFF
--- a/gbif/pipelines/ingest-gbif-beam/src/main/resources/elasticsearch/es-occurrence-schema.json
+++ b/gbif/pipelines/ingest-gbif-beam/src/main/resources/elasticsearch/es-occurrence-schema.json
@@ -157,7 +157,7 @@
         }
       },
       "occurrenceStatus": {"type": "keyword", "copy_to": "all"},
-  
+
       "organismQuantity": {"type": "double"},
       "organismQuantityType": {"type": "keyword", "copy_to": "all"},
       "sampleSizeUnit": {"type": "keyword", "copy_to": "all"},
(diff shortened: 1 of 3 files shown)


gbif/pipelines/ingest-gbif-beam/src/main/resources/elasticsearch/es-occurrence-schema.json
```diff
@@ -203,6 +203,7 @@
       "gadm": {
         "type": "object",
         "properties": {
+          "gid": {"type": "keyword"},
```
timrobertson100, line 206: Looks like we use plural naming convention in our schema. `gids`  ?

MattBlissett: I wasn't sure, as we have `taxonKey` which is singular.

muttcg (approved):
LGTM
Travis says:
Run 'mvn spotless:apply' to fix these violations.

MattBlissett:
I didn't have any warnings running it locally, but I've rebased it anyway.

Do we want `gid` or `gids`?

muttcg:
> I didn't have any warnings running it locally, but I've rebased it anyway.
> 
> Do we want `gid` or `gids`?

It is up to you, I am not so sensitive to that kind of things

MattBlissett:
I have changed it to `gids`, which is more consistent with `identifiedByIds` etc, leaving `taxonKey` as the exception.

PR #393, opened 2020-09-17
388 separate date parser

muttcg:
This is a draft PR
I want to do next:
1) Get rid of default day and month in interpreted avro
2) EU/US format settings
3) Date ranges

This PR covers 1 topic, I removed only the default day and month, I haven't applied formatting anywhere else (For GBIF it should be GbifJsonConverter and HDFS convert or occurrence-ws level)

For 2 and 3 I would like to ask Bai to help me, cause he has done a great job, @qifeng-bai I know you already created PR, but from ala-dev, cause we have a big difference between ala-dev and gbif-dev, could you create a branch from 388_separate_date_parser and migrate "EU/US format settings" first, and after create PR to 388_separate_date_parser where we can discuss changes when PR will be merged into 388_separate_date_parser, we can do the same with "Date ranges"

The goal is to improve dates interpretation in master first and after I will start with merging ala-dev with the latest master


qifeng-bai:
I created two PRs: [388 eu us date](https://github.com/gbif/pipelines/pull/394) and [388_date_range](https://github.com/gbif/pipelines/pull/395)

PR #394, opened 2020-09-18
Migrate eu_au date parsing to 388

qifeng-bai:
@muttcg  

In this solution, we set a **static** TemporalInterpreter based on config.  It should be safe, because an interpreter runs on an independent VM.  However, in our interpreter test case, we need to make sure "reset" the temporalInterpreter to the original one after every test, in case we set the temporal interpreter to 'DMY' or 'MYD'.

I am thinking if we make current 'static' TemporalInterpreter to a 'none-static' class, and initiate an instance based on config, would be safer and better


muttcg:
Thanks, Bai. Here is though:
I think we must move "config" method from TemporalInterpreter.java to another level, the reason it that config is static and at GBIF we would like to pass the date format via an interpretation option because we can change it dynamically. 

My vision is:

1) Leave parameter in PipelinesConfig and add new to InterpretationPipelineOptions (useful for gbif)
2) Get rid "static" from TemporalInterpreter, we run java version of pipelines in multithread, can be many datasets using the same static object, and they mustn't affect each other, in the current realization one dataset can turn on the wrong date parser for another thread (possible bug)
3) Create util method/class  -> TemporalParser createTemporalParser(DateComponentOrdering enum):
4) Use DateComponentOrdering enum as a parameter for TemporalTransform constructor, and init TemporalParser temporalParser, pass temporalParser to TemporalInterpreter (for performance reason, otherwise we will init it in TemporalInterpreter every time)
5) Use DateComponentOrdering enum and TemporalParser temporalParser as a parameter for TemporalInterpreter constructor, and use Lombok @Builder for private constructor, if temporalParser is null, use enum and try to init temporalParser using the util from step 3 (for easier unit testing)

muttcg (changes requested):
Thanks, Bai. Here is though:
I think we must move "config" method from TemporalInterpreter.java to another level, the reason it that config is static and at GBIF we would like to pass the date format via an interpretation option because we can change it dynamically. 

My vision is:

1) Leave parameter in PipelinesConfig and add new to InterpretationPipelineOptions (useful for gbif)
2) Get rid "static" from TemporalInterpreter, we run java version of pipelines in multithread, can be many datasets using the same static object, and they mustn't affect each other, in the current realization one dataset can turn on the wrong date parser for another thread (possible bug)
3) Create util method/class  -> TemporalParser createTemporalParser(DateComponentOrdering enum):
4) Use DateComponentOrdering enum as a parameter for TemporalTransform constructor, and init TemporalParser temporalParser, pass temporalParser to TemporalInterpreter (for performance reason, otherwise we will init it in TemporalInterpreter every time)
5) Use DateComponentOrdering enum and TemporalParser temporalParser as a parameter for TemporalInterpreter constructor, and use Lombok @Builder for private constructor, if temporalParser is null, use enum and try to init temporalParser using the util from step 3 (for easier unit testing)

qifeng-bai:
2, 4 were original implemented I did, I see if I can easily rollback. @muttcg

muttcg (approved):

PR #395, opened 2020-09-18
388 date range

qifeng-bai:
@muttcg  388_date_range was branched from 388_eu_au_date.

It calls TemporalAccessorUtils.toEarliestLocalDateTime / TemporalAccessorUtils.toLatestLocalDateTime before set EventDate.LTE and EventDate.GTE
For example:
1990-01-01/15 will generate eventDate.GTE = 1990-01-10T00:00.   eventDate.LTE = 1990-01-15T23:59

 It conflicts with you current solution which  sets LTE/GTE only to "Day", not to "Time".


sdks/core/src/main/java/org/gbif/pipelines/core/interpreters/core/TemporalInterpreter.java
```diff
@@ -41,80 +48,276 @@
...
+    //          .map(AtomizedLocalDate::fromTemporalAccessor)
+    //          .ifPresent(
+    //              ald -> {
+    //                tr.setYear(ald.getYear());
+    //                tr.setMonth(ald.getMonth());
+    //                tr.setDay(ald.getDay());
+    //              });
+    //    }
+    //    addIssueSet(tr, eventResult.getIssues());
+  }
```
muttcg, line 121: Please, don't leave commented code, just delete, if we need it, we can restore using git history

muttcg (changes requested):
Thanks for the changes, Bai
I think we must refactor **TemporalInterpreter** class and leave only 3 methods:

- interpretTemporal
- interpretModified
- interpretDateIdentified

Others must be in the **parsers** package, create a new class, please. The initial idea for core module packages is:

- converters (convert from one type object to another, get -> set, or similar)
- interpreters (high level where we have avro models and we use classes from parsers to parse, and set fields in avro models)
- parsers (main parsers where we don't use avro object models, example)

It should be a parser class that takes some strings values and returns object similar/or to org.gbif.pipelines.core.parsers.temporal.ParsedTemporal and after we use it in TemporalInterpreter to set fields in TemporalRecord:

example:
`
interpretTemporal(ExtendedRecord er, TemporalRecord tr){
....
ParsedTemporal pt = parseDateRange(String year, String month, String day, String dateString);
pt.getYearOpt.ifPresent(tr::setYear)
..etc
}

`

sdks/core/src/main/java/org/gbif/pipelines/core/interpreters/core/TemporalInterpreter.java
```diff
@@ -41,80 +48,276 @@
...
+
+    boolean atomizedDateProvided =
+        StringUtils.isNotBlank(year)
+            || StringUtils.isNotBlank(month)
+            || StringUtils.isNotBlank(day);
+    boolean dateStringProvided = StringUtils.isNotBlank(eventDateString);
+
+    if (atomizedDateProvided || dateStringProvided) {
+      TemporalAccessor parsedStartTemporalAccessor;
+      ParseResult.CONFIDENCE confidence;
```
muttcg, line 179: It is not used in the code

sdks/core/src/main/java/org/gbif/pipelines/core/interpreters/core/TemporalInterpreter.java
```diff
@@ -41,80 +48,276 @@
...
+          }
+        }
+
+        // Solve conflicts of dates from YMD and dateString
+        if (atomizedDateProvided
+            && dateStringProvided
+            && !TemporalAccessorUtils.sameOrContained(parsedYMDResult.getPayload(), from)) {
+          // eventDate could be ambiguous (5/4/2014), but disambiguated by year-month-day.
+          boolean ambiguityResolved = false;
+          if (startParseResult.getAlternativePayloads() != null) {
```
muttcg, line 225: Possible NPE, startParseResult can be null

sdks/core/src/main/java/org/gbif/pipelines/core/interpreters/core/TemporalInterpreter.java
```diff
@@ -41,80 +48,276 @@
...
+        Optional.ofNullable(parsed.getPayload())
+            .map(TemporalAccessor::toString)
             .ifPresent(tr::setDateIdentified);
       }
+
       addIssueSet(tr, parsed.getIssues());
     }
   }
 
+  private static void interpretEventDate(ExtendedRecord er, TemporalRecord tr) {
```
muttcg, line 161: Try to refactor this method it is too big, split it into several small

sdks/core/src/main/java/org/gbif/pipelines/core/interpreters/core/TemporalInterpreter.java
qifeng-bai: @muttcg  Thanks for your advice. I forgot to remove the comments.  

If there are only three methods should be in, then interpretTemporal should be called interpretEventDate, I think

I have been considering a general range parser,  I found it could be hard, because: for startDate, we need provided year, month, day to solve ambiguous date, mainly on like 01/02/2004.
but for end date, we don't need to that.

Like what you mentioned, passing year, month, day into the method, looks not very generic. 

ParsedTemporal pt = parseDateRange(String year, String month, String day, String dateString);

I am still working on it

PR #403, opened 2020-10-13
369 grscicoll improvements

marcos-lg:
Removed unnecesary fields in collections since they are not used and API changes can break pipelines.


muttcg (approved):
LGTM

PR #407, opened 2020-10-21
329 lifestage vocabulary

marcos-lg:
Contains this other PR https://github.com/gbif/pipelines/pull/313 updated with the latest changes in gbif-dev

### DIFF
--- a/sdks/core/src/main/java/org/gbif/pipelines/core/parsers/VocabularyParser.java
+++ b/sdks/core/src/main/java/org/gbif/pipelines/core/parsers/VocabularyParser.java
@@ -59,6 +59,7 @@ public class VocabularyParser<T extends Enum<T>> {
   }
 
   /** @return a life stage parser. */
+  @Deprecated
   public static VocabularyParser<LifeStage> lifeStageParser() {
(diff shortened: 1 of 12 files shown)


sdks/core/src/main/java/org/gbif/pipelines/core/parsers/VocabularyParser.java
```diff
@@ -59,6 +59,7 @@
   }
 
   /** @return a life stage parser. */
+  @Deprecated
   public static VocabularyParser<LifeStage> lifeStageParser() {
     return new VocabularyParser<>(LST_PARSER, DwcTerm.lifeStage);
   }
```
muttcg, line 65: As I understand we don't use it at all, maybe just delete?

muttcg (approved):
Thanks, LGTM

sdks/core/src/main/java/org/gbif/pipelines/core/parsers/VocabularyParser.java
marcos-lg: right, I just deleted it

muttcg (approved):

PR #411, opened 2020-10-23
Improved dwca->avor performance

muttcg:
Removed stream filter and use Record if an extension is empty to improve performance, overall it gives 30% performance boost


sdks/core/src/main/java/org/gbif/pipelines/core/converters/ExtendedRecordConverter.java
```diff
@@ -22,25 +24,33 @@
   private static final Function<Record, Map<String, String>> REMOVE_EMPTY_CONTENT =
       record ->
           record.terms().stream()
-              .filter(term -> term.qualifiedName() != null && record.value(term) != null)
-              .collect(Collectors.toMap(Term::qualifiedName, record::value, (a, b) -> b));
+              .collect(
+                  HashMap::new,
```
timrobertson100, line 28: I see you removed the sizing here (could be sized on record.terms()/2 or so)

fmendezh (approved):

PR #442, opened 2020-12-04
renamed isClustered to isInCluster

marcos-lg:
The term for the cluster field was named as isInCluster (https://github.com/gbif/dwc-api/pull/5) so the avro schema of the hdfs view has to match that name to create the Hive table (https://github.com/gbif/gbif-api/pull/77).

Maybe we should change the name in ES and in the BasicRecord too.


muttcg:
@timrobertson100 Do we need to change the name of the parameter in ES schema then?

muttcg (approved):

PR #461, opened 2021-01-19
16 jackknife

adam-collins:


sdks/core/src/main/java/org/gbif/pipelines/core/parsers/location/cache/GeocodeBitmapCache.java
```diff
@@ -85,15 +85,15 @@ private GeocodeResponse getDefaultGeocodeResponse(
     locations = loadFn.apply(LatLng.builder().withLatitude(lat).withLongitude(lng).build());
     // Don't store this if there aren't any locations.
     if (locations.getLocations().isEmpty()) {
-      log.error(
+      log.debug(
```
djtfmartin, line 88: Could we revert the changes to logging level ? 
This may affect GBIF.

djtfmartin (changes requested):
Looks good. The only this is to revert the logging level changes. Otherwise, it can be merged.

PR #467, opened 2021-01-25
15 duplicate detection - clustering port from occurrence-clustering to beam

djtfmartin:
This is a beam port of GBIF's clustering which now makes use of the same classes used in GBIF spark job.

Related issue: https://github.com/AtlasOfLivingAustralia/la-pipelines/issues/15

### DIFF
--- a/livingatlas/pipelines/src/main/java/au/org/ala/clustering/HashKeyOccurrenceBuilder.java
+++ b/livingatlas/pipelines/src/main/java/au/org/ala/clustering/HashKeyOccurrenceBuilder.java
@@ -1,0 +1,162 @@
+package au.org.ala.clustering;
+
+public final class HashKeyOccurrenceBuilder {
+  String hashKey = null;
+  String id = null;
+  String datasetKey = null;
+  String speciesKey = null;
+  String taxonKey = null;
+  String basisOfRecord = null;
+  Double decimalLatitude = null;
+  Double decimalLongitude = null;
+  Integer year = null;
+  Integer month = null;
+  Integer day = null;
+  String eventDate = null;
+  String scientificName = null;
+  String countryCode = null;
+  String typeStatus = null;
+  String occurrenceID = null;
+  String recordedBy = null;
+  String fieldNumber = null;
+  String recordNumber = null;
+  String catalogNumber = null;
+  String otherCatalogNumbers = null;
+
+  private HashKeyOccurrenceBuilder() {}
+
+  public static HashKeyOccurrenceBuilder aHashKeyOccurrence() {
+    return new HashKeyOccurrenceBuilder();
+  }
+
+  public HashKeyOccurrenceBuilder withHashKey(String hashKey) {
+    this.hashKey = hashKey;
+    return this;
+  }
+
+  public HashKeyOccurrenceBuilder withId(String id) {
+    this.id = id;
+    return this;
+  }
+
+  public HashKeyOccurrenceBuilder withDatasetKey(String datasetKey) {
+    this.datasetKey = datasetKey;
+    return this;
+  }
+
+  public HashKeyOccurrenceBuilder withSpeciesKey(String speciesKey) {
+    this.speciesKey = speciesKey;
+    return this;
+  }
+
+  public HashKeyOccurrenceBuilder withTaxonKey(String taxonKey) {
+    this.taxonKey = taxonKey;
+    return this;
+  }
+
+  public HashKeyOccurrenceBuilder withBasisOfRecord(String basisOfRecord) {
+    this.basisOfRecord = basisOfRecord;
+    return this;
+  }
+
+  public HashKeyOccurrenceBuilder withDecimalLatitude(Double decimalLatitude) {
+    this.decimalLatitude = decimalLatitude;
+    return this;
+  }
+
+  public HashKeyOccurrenceBuilder withDecimalLongitude(Double decimalLongitude) {
+    this.decimalLongitude = decimalLongitude;
+    return this;
+  }
+
+  public HashKeyOccurrenceBuilder withYear(Integer year) {
+    this.year = year;
+    return this;
+  }
+
+  public HashKeyOccurrenceBuilder withMonth(Integer month) {
+    this.month = month;
+    return this;
+  }
+
+  public HashKeyOccurrenceBuilder withDay(Integer day) {
+    this.day = day;
+    return this;
+  }
+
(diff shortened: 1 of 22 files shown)


livingatlas/pipelines/src/main/java/au/org/ala/clustering/OccurrenceRelationships.java
```diff
@@ -0,0 +1,251 @@
+package au.org.ala.clustering;
```
timrobertson100, line 1: I foresee this will quickly become a mess if we fork code like this. These are slightly modified versions of the GBIF codebase and will deviate.

Wouldn't it be better to have common code to work on?

djtfmartin: Fully agree. Not sure how best to proceed without refactoring the original `occurrence-clustering` module, removing the use of Spark artefacts.

First thought is that `OccurrenceRelationships` logic should be dealing with POJOs without Spark, AVRO etc. and perhaps this should be moved out to a separate library.

timrobertson100:
Once https://github.com/gbif/pipelines/pull/486 is merged I think this can continue.

I suggest:

1. a rebase since this is old and many changes have come in
2. consider moving the `OccurrenceFeaturesPojo` from the sdk/core/src/test into the src/main if it suits your needs rather than transient Avro objects

livingatlas/pipelines/src/main/java/au/org/ala/clustering/HashKeyOccurrenceBuilder.java
```diff
@@ -0,0 +1,162 @@
+package au.org.ala.clustering;
+
+public final class HashKeyOccurrenceBuilder {
```
timrobertson100, line 3: Just curious - why not Lombok `@Builder` on the HashKeyOccurrence?

livingatlas/pipelines/src/main/java/au/org/ala/clustering/RepresentativeRecordUtils.java
```diff
@@ -0,0 +1,187 @@
...
+    if (highestRanking.size() == 1) {
+      return highestRanking.get(0);
+    }
+
+    // get highest resolution dates
+    highestRanking = rankDatePrecision(highestRanking);
+    if (highestRanking.size() == 1) {
+      return highestRanking.get(0);
+    }
+    if (highestRanking.size() == 1) {
```
timrobertson100, line 30: Something not right here (cut 'n paste mistake?)

livingatlas/pipelines/src/main/java/au/org/ala/clustering/RepresentativeRecordUtils.java
```diff
@@ -0,0 +1,187 @@
...
+        new Comparator<HashKeyOccurrence>() {
+          @Override
+          public int compare(HashKeyOccurrence o1, HashKeyOccurrence o2) {
+            return o1.getId().compareTo(o2.getId());
+          }
+        });
+    return cluster.get(0);
+  }
+
+  public static List<HashKeyOccurrence> rankDatePrecision(List<HashKeyOccurrence> cluster) {
```
timrobertson100, line 49: Suggest add a jdoc

livingatlas/pipelines/src/main/java/au/org/ala/clustering/RepresentativeRecordUtils.java
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    if (occurrenceFeatures.getDay() != null) {
+      datePrecision += 1;
+    }
+
+    return datePrecision;
+  }
+
+  public static List<List<HashKeyOccurrence>> createClusters(List<ClusterPair> pairs) {
```
timrobertson100, line 160: Suggest add clusters

livingatlas/pipelines/src/main/java/au/org/ala/pipelines/beam/ClusteringPipeline.java
```diff
@@ -0,0 +1,445 @@
...
+import org.gbif.pipelines.core.parsers.clustering.OccurrenceFeatures;
+import org.gbif.pipelines.core.parsers.clustering.OccurrenceRelationships;
+import org.gbif.pipelines.core.parsers.clustering.RelationshipAssertion;
+import org.gbif.pipelines.core.utils.FsUtils;
+import org.gbif.pipelines.io.avro.*;
+import org.slf4j.MDC;
+
+/**
+ * Clustering pipeline which is a Apache Beam port of the work in the module
+ * https://github.com/gbif/occurrence/tree/master/occurrence-clustering
```
timrobertson100, line 34: Please remove this sentence

livingatlas/solr/conf/managed-schema
```diff
@@ -361,6 +361,16 @@
   <field name="waterBody"  type="string" docValues="true" indexed="true"/>
   <field name="year"  type="int" docValues="true" indexed="true"/>
 
+  <!-- Clustering (Duplicate detection) -->
+  <field name="isClustered" type="boolean" docValues="true" indexed="true"/>
+  <field name="clusteredWith" type="string" docValues="true" indexed="true" multiValued="true"/>
```
timrobertson100, line 366: Be aware that I found examples where we have 18000+ clustering records due to the nature of data. In that example it was gut analysis from intestines, where the metadata was all about the host organism collections, so all the gut samples clustered with everything at the same location.

timrobertson100 (approved):
There are a few tiny changes to make (one that looks like a copy/paste error but might need investigation by you). Once addressed, please just self merge.

Thanks for accommodating the wish to align the GBIF and ALA code properly, which delayed this by a month.

livingatlas/pipelines/src/main/java/au/org/ala/clustering/HashKeyOccurrenceBuilder.java
djtfmartin: Problems/conflicts with the Apache Beam JavaBean schema implementation. Sorry - I didn't make a note of the errors at the time, but I couldn't get it to work with lombok.

livingatlas/solr/conf/managed-schema
djtfmartin: Thanks for this. Ive added a configurable `candidatesCutoff` which is used by the pipeline to drop a cluster of occurrences over a certain size. This is defaults to 50.

PR #469, opened 2021-01-26
465 remove lookups

muttcg:
Feel free to improve my _RuNglish_


gbif/diagnostics/src/main/java/org/gbif/pipelines/diagnostics/RepairGbifIDLookupTool.java
```diff
@@ -0,0 +1,187 @@
...
+import org.gbif.pipelines.keygen.HBaseLockingKeyService;
+import org.gbif.pipelines.keygen.common.HbaseConnectionFactory;
+import org.gbif.pipelines.keygen.config.KeygenConfig;
+import org.gbif.pipelines.keygen.identifier.OccurrenceKeyBuilder;
+
+@Slf4j
+@Builder
+public class RepairGbifIDLookupTool {
+
+  @Parameter(names = "--datase-key", description = "GBIF registry ID for the dataset")
```
timrobertson100, line 26: typo --datase[t]-key

muttcg: Indeed, thanks

gbif/diagnostics/src/main/java/org/gbif/pipelines/diagnostics/RepairGbifIDLookupTool.java
```diff
@@ -0,0 +1,187 @@
...
+  @NotNull
+  public String datasetKey;
+
+  @Parameter(
+      names = "--input-directory",
+      description = "Use DWCA archive as a lookup keys source, provide full input path")
+  public File dwcaDirectory;
```
timrobertson100, line 33: Should this be named dwcaSource? It seems to support both files and directories.

timrobertson100 (commented):
Am I correct in thinking that since this uses a DwC-A as a source, it still has the potential to leave lookups that aren't used in the ES index? That might be a different tool, but I still expect that would be the most commonly used approach (delete everything that is not related to the latest attempt) which would ensure strong consistency.

muttcg:
> Am I correct in thinking that since this uses a DwC-A as a source, it still has the potential to leave lookups that aren't used in the ES index? That might be a different tool, but I still expect that would be the most commonly used approach (delete everything that is not related to the latest attempt) which would ensure strong consistency.

We can use dwca, the logic is:
1) full scan by datasetKey prefix, create set with lookup keys
2) get lookup from dwca records and remove all lookup keys from the set
3) use the result set and delete records from hbase

Issue #470, opened 2021-01-28
Remove use of ala-sensitive-data-service snapshot from build


djtfmartin:
The `livingatlas/pipelines` module is currently referencing a SNAPSHOT version of ala-sensitive-data-service-client.

We need to have a release version of this library in nexus.ala.org.au (currently we only have snapshots) and reference in POM.

charvolant:
Requires non-snapshot versions of ala-name-matching and sds, as well before build.

djtfmartin:
Done. Thanks @charvolant

djtfmartin:
@charvolant  - Not done :)

The build for `ala-sensitive-data-service` is failing at the last stage (the upload to nexus.ala.org.au).

https://travis-ci.com/github/AtlasOfLivingAustralia/ala-sensitive-data-service

charvolant:
Partially caused by ALA nexus having a cow when there is a -test.jar artefact in releases. The maven deploy attempts to update the .pom twice and breaks the nexus release rules. Fixed in https://github.com/AtlasOfLivingAustralia/ala-sensitive-data-service/pull/7

Why it's throwing a 400 error after it's successfully uploaded the sources is still under investigation.

vjrj:
We are getting this error trying to build the dev branch
```
[ERROR] Failed to execute goal on project pipelines: Could not resolve dependencies for project au.org.ala:pipelines:jar:2.8.0-SNAPSHOT: Failed to collect dependencies at au.org.ala.sds:ala-sensitive-data-client:jar:1.1-SNAPSHOT: Failed to read artifact descriptor for au.org.ala.sds:ala-sensitive-data-client:jar:1.1-SNAPSHOT: Could not find artifact au.org.ala.sds:ala-sensitive-data-service:pom:1.1-SNAPSHOT in gbif-all (http://repository.gbif.org/content/groups/gbif) -> [Help 1]
```

I noticed that https://nexus.ala.org.au/content/repositories/snapshots/au/org/ala/sds/ala-sensitive-data-client/1.1-SNAPSHOT/maven-metadata.xml gives `404`.

Although the nexus version does not match sounds like:
https://issues.sonatype.org/browse/NEXUS-24988

With `1.0-SNAPSHOT` downloads well, but I get some compilation errors.

TIA

djtfmartin:
Ive fixed this now. 1.1 is now in the ALA nexus repository. This can be closed.

PR #476, opened 2021-02-11
Stable versions of sensitive data service

charvolant:
Also updates docker images with versions with the stable service.


charvolant:
Fix for #470

PR #477, opened 2021-02-11
272 vertnet features

muttcg:


sdks/core/src/main/java/org/gbif/pipelines/core/parsers/dynamic/DynamicParser.java
```diff
@@ -0,0 +1,119 @@
...
+      result = result.replace(entry.getKey(), entry.getValue());
+    }
+    mainTemplateMap.put(key, result);
+  }
+
+  protected void initPattern(String value) {
+    initPattern(value, null, null);
+  }
+
+  protected void initPattern(String value, String defaultKey, String defaultType) {
```
fmendezh, line 33: since the elements added here are mostly RegEx, shouldn't the compiled version be stored?

sdks/core/src/main/java/org/gbif/pipelines/core/parsers/dynamic/DynamicParser.java
```diff
@@ -0,0 +1,119 @@
...
+    String result = value;
+    for (Map.Entry<String, String> entry : mainTemplateMap.entrySet()) {
```
fmendezh, line 35: let's document somewhere the expected behavior of this, it looks to me that is not initiating anything but checking that a pattern is not replacing a default pattern

sdks/core/src/main/java/org/gbif/pipelines/core/parsers/dynamic/DynamicParser.java
```diff
@@ -0,0 +1,119 @@
...
+          result = s;
+        } else {
+          result = String.join(", ", result, s);
+        }
+      }
+    }
+    return result == null ? value : result;
+  }
+
+  protected void initTemplates() {
```
fmendezh, line 64: not really sure what is happening here...it is keeping the last value of a multimap?...not sure if that should be called init?

fmendezh (changes requested):
All looks good except for the internal implementation of DynamicParser which looks very confusing in terms of logic and naming conventions

muttcg:
About **sdks/core/src/main/java/org/gbif/pipelines/core/parsers/dynamic/DynamicParser.java**
All parsers are copies of vertnet python parsers, as I understood (maybe I am wrong) python (or some libs) support regex templates, so it is a cycled graph of regex templates with default values for regex groups:

Example:
basic templates are **not** applied directly to a value:
```
x = "hello"
y = "%x world"
z = "%x %y"
```

parser templates are applied directly to a value:
```
xx = "%x %z something"
xxy = "%xx something %x"
```

when we compile regex, we get something like:
```
xx = "hello hello hello world something"
xxy = "hello hello hello world something something hello"
```
Also if regex matcher can be applied, but group<key> is null, then will be used default value

Compiled regex in the end:
`\b(?<key>total[\s\-]?length[\s\-]?in[\s\-]?mm|total[\s\-]?length[\s\-]?in|length[\s\-]?in[\s\-]?millimeters|(?:total|max|standard)[\s\-]?lengths?|meas(?:[a-zA-Z]*)?\.?:\s*l|s\.?l\.?|label\.?\s*lengths?|snout[\s\-]?vent[\s\-]?lengths?(?:[\s\-]?in[\s\-]?mm)?|s\.?v\.?(:?l\.?)?|snout\s+vent\s+lengths?|head[\s\-]?body[\s\-]?length[\s\-]?in[\s\-]?millimeters|(?:fork|mean|body)[\s\-]?lengths?|t[o.]?l\.?_?|lengths?|tag|measurements?|body|total|on\s*tag|specimens?|catalog|measurements(?:\s+[a-zA-Z]+)|tag\s+\d+\s*=?(?:male|female)?\s*,|meas[.,]?(?:\s+\w+\.\w+\.)?|mesurements|measurementsnt)?\s*[^\w.\[\(]*\s*?(?<value1>[\(\[\{]?(?:\d{1,3}(?:,\d{3}){1,3}|\d+)(?:\.\d+)?[\)\]\}]?[\*]?(?:\s*(?:-|to)\s*[\(\[\{]?(?:\d{1,3}(?:,\d{3}){1,3}|\d+)(?:\.\d+)?[\)\]\}]?[\*]?)?)\s*(?<units1>(?:foot|feet|ft)s?\.?)\s*(?<value2>[\(\[\{]?(?:\d{1,3}(?:,\d{3}){1,3}|\d+)(?:\.\d+)?[\)\]\}]?[\*]?(?:\s*(?:-|to)\s*[\(\[\{]?(?:\d{1,3}(?:,\d{3}){1,3}|\d+)(?:\.\d+)?[\)\]\}]?[\*]?)?)\s*(?<units2>(?:inche?|in)s?\.?)`

And there are 14 regexes like above applied in the loop, the first match wins. So I basically ported the same behavior to java

**sdks/core/src/main/java/org/gbif/pipelines/core/parsers/dynamic/DynamicParser.java** is package only and used inside 
**org.gbif.pipelines.core.parsers.dynamic.MassParser** and **org.gbif.pipelines.core.parsers.dynamic.LengthParser**, those parsers are static and all regex functions will be assembled from templates and compiled during class-loading time

Issue #498, opened 2021-03-17
Update the GRSciColl matching


ManonGros:
We need a few changes to the GRSciColl matching:

- Don't link occurrences to collections if we have "institution match none"
- If record = specimen*
    - link it or flag it if collection, institution code and ID are given
- if record != specimen
    -  link it if possible and not ambiguous, otherwise silently ignore it (in other words, don't flag the records)

\* specimen include: PRESERVED_SPECIMEN, FOSSIL_SPECIMEN, LIVING_SPECIMEN, MATERIAL_SAMPLE

marcos-lg:
Deployed to PROD. The changes will be visible when the datasets are reinterpreted.

marcos-lg:
Deployed to PROD. Changes in occurrences will be visible when the cache is refreshed.

albenson-usgs:
I'm still seeing flags for records that are HumanObservation. https://www.gbif.org/occurrence/1847197337

ManonGros:
Thanks for letting us know @albenson-usgs this is because the data wasn't reinterpreted. It should be fixed now but let us know if you find any other.

PR #501, opened 2021-03-18
#498 grscicoll matching

marcos-lg:
This is still a work in progress but I want it to check it with you already before going further.

I had to change the grscicoll transform to receive as input both the extendedRecord and the basicRecord since I have to use the basis of record in the grscicoll interpreter.

I've put a ToDo in a couple of places where I had to do a workaround. To avoid this workaround I was thinking to change the `check` method in the `Transform` class to sth like this:

```
  public CheckTransforms<R> check(Set<String> types) {
    return CheckTransforms.create(
        classInput, CheckTransforms.checkRecordType(types, recordType));
  }
```
Being `classInput` a `Class<R>` that should be passed in the constructor so we'd have to change all transforms.

What do you think? any other ideas?

(I haven't run any pipeline with this code yet so there might be issues)

### DIFF
--- a/sdks/beam-transforms/src/main/java/org/gbif/pipelines/transforms/core/GrscicollTransform.java
+++ b/sdks/beam-transforms/src/main/java/org/gbif/pipelines/transforms/core/GrscicollTransform.java
@@ -7,14 +7,18 @@ import java.io.IOException;
 import java.time.Instant;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.Set;
 import lombok.Builder;
+import lombok.NonNull;
 import lombok.Setter;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.beam.sdk.transforms.MapElements;
 import org.apache.beam.sdk.transforms.ParDo;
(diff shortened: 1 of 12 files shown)


sdks/beam-transforms/src/main/java/org/gbif/pipelines/transforms/core/GrscicollTransform.java
```diff
@@ -96,21 +107,31 @@ public void tearDown() {
...
   public void processElement(ProcessContext c) {
-    processElement(c.element(), c.sideInput(metadataView)).ifPresent(c::output);
+    CoGbkResult v = c.element();
+
+    ExtendedRecord er = v.getOnly(erTag, null);
+    BasicRecord br = v.getOnly(brTag, null);
+
+    if (er == null || br == null) {
+      return;
+    }
```
muttcg, line 124: Should this check be in processElement? If so then it will work for beam and java versions

muttcg (commented):
It would be nice if you add a test for GrscicollTransform.java and also add javadoc comments for some new methods

muttcg (approved):
Thanks, looks good!

PR #502, opened 2021-03-18
277 incremental sampling

djtfmartin:
Changes to allow sampling for only the newly added occurrences with new coordinates.

See: https://github.com/AtlasOfLivingAustralia/la-pipelines/issues/277


livingatlas/pipelines/src/main/java/au/org/ala/pipelines/beam/SamplingPipeline.java
```diff
@@ -0,0 +1,232 @@
...
+  public static void run(SamplingPipelineOptions options) throws Exception {
+
+    FileSystem fs =
+        FileSystemFactory.getInstance(options.getHdfsSiteConfig(), options.getCoreSiteConfig())
+            .getFs(options.getTargetPath());
+
+    log.info("Checking for new layers in the system");
+    SamplingService samplingService = SamplingUtils.initSamplingService(options.getBaseUrl());
+
+    boolean newLayersAvailable = newLayersAddedSinceLastSample(samplingService, options, fs);
```
sadeghim, line 70: Should we check for dropped layers too? If a layer been deleted, do we need to remove that from the sampling?

sadeghim (commented):
All good,
Just a question and a suggestion in the review comments.

livingatlas/pipelines/src/main/java/au/org/ala/pipelines/beam/SamplingPipeline.java
djtfmartin: Thanks @sadeghim. I agree it would be nice to handle this. Ive raised this issue https://github.com/AtlasOfLivingAustralia/la-pipelines/issues/290 so we can handle it in future PR.

livingatlas/pipelines/src/main/java/au/org/ala/pipelines/beam/SamplingPipeline.java
```diff
@@ -0,0 +1,232 @@
...
+  }
+
+  /**
+   * Load index records from AVRO.
+   *
+   * @param options
+   * @param p
+   * @return
+   */
+  private static PCollection<IndexRecord> loadIndexRecords(
```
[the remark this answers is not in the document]
djtfmartin: @sadeghim refactored to ALAFsUtils and implemented as suggested

Issue #521, opened 2021-05-07
Catalogue of Life taxonomy comparison (Phase I)


timrobertson100:
The pipeline for generating the pre-release comparison for the backbone should be adapted to generate a report for what the data would look like according to a COL release.

The output of this will be presented in the COL data management console for data managers to view and provide feedback for enhancements. 

Future phases will introduce a second taxonomy backbone to GBIF.

PR #526, opened 2021-05-19
Changes required to satisfy COL taxonomy comparison #521

timrobertson100:
This brings in changes so that the keys can be ignored in the comparison, a backend API can be used (i.e. not using the /v1/...) which allows first example report to be generated for the Catalogue of Life. There are examples of those reports on https://github.com/CatalogueOfLife/checklistbank/issues/877.


fmendezh (changes requested):
BackbonePreReleaseOptions has default values to UAT servers and Hive databases, consider removing them and adding those as examples in the README file

timrobertson100:
Thanks @fmendezh  https://github.com/gbif/pipelines/pull/526/commits/938d9540b0c386372a1a229859cc1143c3e3e319

fmendezh (approved):

PR #532, opened 2021-05-28
override NULL licence at record level with licence at collectory level

adam-collins:


djtfmartin (approved):

livingatlas/pipelines/src/main/java/au/org/ala/pipelines/transforms/IndexRecordTransform.java
```diff
@@ -404,9 +405,15 @@ public static IndexRecord createIndexRecord(
...
+      // or licence is License.UNSPECIFIED in the basic record (licence provided at the record-level
+      // is null)
       // then use the licence supplied by the collectory
       // see https://github.com/AtlasOfLivingAustralia/la-pipelines/issues/271
-      if (indexRecord.getStrings().get(DcTerm.license.simpleName()) == null) {
+      if (indexRecord.getStrings().get(DcTerm.license.simpleName()) == null
+          || indexRecord
+              .getStrings()
+              .get(DcTerm.license.simpleName())
+              .equals(License.UNSPECIFIED.name())) {
```
timrobertson100, line 416: It might be worth adding a check for `License.UNSUPPORTED` here as well. The difference being unspecified is missing or unrecognisable, while unsupported is recognised but not accepted (i.e. not CC0, CC-BY or CC-BY-NC)

PR #536, opened 2021-06-03
add SOLR schema textgen searchable fields for identifiedBy and recordedBy

adam-collins:


brucehyslop (approved):

djtfmartin:
Can we chose a naming convention that isn't so tied to a specific field type in SOLR?

djtfmartin (changes requested):
Can we choose a naming convention that isn't so tied to a specific field type in SOLR?

brucehyslop:
Is there a reason why we couldn't change the type of the fields `recordedBy` and `identifiedBy` from `text` to `textgen`? 

Would this have a detrimental effect when searching on these fields?

djtfmartin:
Closing this PR after discussion with @adam-collins 

We've renamed the fields to `text_recordedBy` and `text_identifiedBy` but keeping textgen field type.

Changes added to this PR: https://github.com/gbif/pipelines/pull/528

PR #547, opened 2021-06-16
Preparations vocab matching

djtfmartin:
Related issue: https://github.com/AtlasOfLivingAustralia/la-pipelines/issues/397

This work creates an interpreted `preparations` field. The main driver for this is to provide a mechanism to filter in/out eDNA records and Genomic DNA records.

The underlying vocab file is under review and is expected to change. Note  - its targeting AU data only at the moment.


m-hope:
> This work creates an interpreted `preparations` field. The main driver for this is to provide a mechanism to filter in/out eDNA records and Genomic DNA records.

Given that filtering out eDNA records is the main driver for this task and given the expected complexity, I would ask (as I have done in issue AtlasOfLivingAustralia/la-pipelines#397), why don't we just use another field that is not currently in use that specifically lists the materialSample type (ie: "Tissue", "EnvironmentalDNA", "GenomicDNA")?

Issue #585, opened 2021-09-10
IndexRecordToDwcaPipeline - archives fail GBIF validation


djtfmartin:
The archives produced by [IndexRecordToDwcaPipeline](https://github.com/gbif/pipelines/blob/dev/livingatlas/pipelines/src/main/java/au/org/ala/pipelines/beam/IndexRecordToDwcaPipeline.java)

More info here: https://github.com/AtlasOfLivingAustralia/la-pipelines/issues/505

There are two identified problems:

1) Invalid meta.xml - problem with the definition of the Multimedia extension

```
  <extension encoding="UTF-8" linesTerminatedBy="\r\n" fieldsTerminatedBy="," fieldsEnclosedBy="&quot;" ignoreHeaderLines="0" rowType="http://rs.gbif.org/terms/1.0/Multimedia">
    <files>
      <location>image.csv</location>
    </files>
    <coreid index="0"/>
    <field index="0" term="id"/>
    <field index="1" term="http://purl.org/dc/terms/identifier"/>
```    

The following line should be removed

```
    <field index="0" term="id"/>    
```    

2) Escaping of fields in the occurrence.csv

Here the file contains JSON in `dynamicProperties` field which breaks the CSV escaping of comma and " delimitation. Similarly fields containing `"AM".` become `""AM"."` which breaks the file.
A workaround for this might be to switch to tab-delimited rather than comma and quotes

timrobertson100:
Examples of errors observed are in the history of https://registry.gbif-uat.org/dataset/ba11ce97-648b-4213-9221-ecab3f0519fc/ingestion-history

djtfmartin:
Work for this is merged now.

PR #588, opened 2021-09-16
Fixes for DwCA export pipeline

djtfmartin:
Changes include:
* Fixes for the export of images
* la-pipelines config changes
* Changes to tab-delimited exports

See: https://github.com/AtlasOfLivingAustralia/la-pipelines/issues/505 and https://github.com/gbif/pipelines/issues/585


livingatlas/pipelines/src/main/java/au/org/ala/pipelines/beam/IndexRecordToDwcaPipeline.java
```diff
@@ -62,18 +65,26 @@ public static void run(SolrPipelineOptions options) {
     indexRecordPCollection
         .apply(
             "Convert to core csv string",
-            MapElements.into(TypeDescriptors.strings()).via(CoreCsvConverter::convert))
+            MapElements.into(strings()).via(CoreCsvConverter::convert))
         .apply(
             "Write core csv file",
```
charvolant, line 70: Don't know if it matters but it might as well be tsv file

livingatlas/pipelines/src/main/java/au/org/ala/pipelines/converters/TsvConverter.java
```diff
@@ -43,10 +43,11 @@ public String converter(T source) {
     keyTermFnList.forEach(
         keyTermFn -> {
           String value = keyTermFn.getFn().apply(source).orElse(keyTermFn.getDefaultValue());
-          value = "\"" + value + "\"";
+          // replace any tabs within string values with single space
+          value = "\"" + value.replaceAll("\t", " ").replaceAll("\"", "") + "\"";
```
charvolant, line 47: Should this properly double-quote escape quotes, rather than deleting them?

charvolant (approved):
I have one concern about quotation marks being deleted, rather than escaped. If that is expected behaviour, then all is good.

livingatlas/pipelines/src/main/java/au/org/ala/pipelines/converters/TsvConverter.java
djtfmartin: I'm intentionally deleting them to hopefully reduce issues for GBIF or other consumers.

Issue #622, opened 2021-11-11
LA-pipelines:  expert distribution outlier detection


qifeng-bai:
Check if a species occurrence record point is in/out of the expert distribution layers

Questions:

1. Where should this occur - sampling or jackknife process in the pipelines or a separate process?
2. Efficiency issue:  Using polygons may slow down the process?  Using Grids? how to determine the size of grids
3. Assertions we should use
4. Possible to Calculate 

> - a distance of the point inside/outside expected distribution field to the record
> - point outside the range and uncertainty overlaps the range
> - point outside the range and uncertainty outside the range

qifeng-bai:
Modification to la-pipeline.yaml is needed 
Add '**outlier**' section :
`outlier:
  appName: Expert distribution outliers for {datasetId}
  baseUrl: https://spatial.ala.org.au/ws/
  inputPath: '{fsPath}/pipelines-outlier'
  targetPath: &outlierTargetPath '{fsPath}/pipelines-outlier'
  allDatasetsInputPath: '{fsPath}/pipelines-all-datasets'
  runner: DirectRunner`

and insert the fellowing to '**solr**' section

`  outlierPath: *outlierTargetPath
    includeOurtlier: false`

qifeng-bai:
SLF4J issues:
1,  found the pipelines  uses logback as the default logger,  la-pipelines script currently is using log4j.properties , logging is not working. 


2, Added logback.xml into ./resource folder.  And import an external config via -Dlogback.configurationFile 

The example of external config looks like:

`<included>`    
`    <logger name="org.gbif" level="info" additivity="false">`
 `       <appender-ref ref="CONSOLE"/>`
  `  </logger>`
   `   <logger name="au.org.ala.pipelines" level="debug" additivity="false">`
    `    <appender-ref ref="CONSOLE"/>`
   ` </logger>      `
`</included>`

We can use it to update the log level.   **It works on my local dev environment**. 

However, we run it on NCI3-spark servers,  the logger does not work. **It does not accept : "included" element.** 

For making it work, we have to **duplicate** the content in ./resource/logback.xml

PR #631, opened 2021-12-02
Calculate distance to expert distribution layers

qifeng-bai:
link to https://github.com/gbif/pipelines/issues/622


livingatlas/pipelines/src/main/java/au/org/ala/distribution/DistributionServiceImpl.java
```diff
@@ -0,0 +1,91 @@
...
+    service = retrofit.create(DistributionService.class);
+  }
+
+  public static DistributionServiceImpl init(String baseUrl) {
+    // set up sampling service
+    return new DistributionServiceImpl(baseUrl);
+  }
+
+  public List<DistributionLayer> getLayers() throws IOException, ExpertDistributionException {
+    // Response<List<DistributionLayer>> response =
```
djtfmartin, line 34: remove commented code

vjrj:
hi @djtfmartin and @qifeng-ba. Testing pipelines I'm getting a new:
```
ERROR File /data/la-pipelines/config/logback.xml doesn't exits
```
This file is neither installed by the debian package or by `ala-install`. Probably the best option is to include it in the debian package. I can add it. 

Any difference between your production and development logback.xml provided in this PR?

PR #657, opened 2022-02-03
Changes to distrbution pipeline

djtfmartin:
* Ensure all outlier records are written to AVRO (not just new ones)
* Remove unused code
* Changed index term to distanceFromExpertDistribution to follow readable camelCase convention
* Use metrics instead of worklog code to write out counts


qifeng-bai (commented):
If writing result to existing input avro file won't cause possible IO conflicts, it should be fine that we write ALL outlier records to file.

Issue #665, opened 2022-02-10
Support multi-value search for existing fields


timrobertson100:
The fields listed below are in ES and treated as single value strings.
Without breaking any pubic APIs, we can provide better search by treating them as multi-value fields.

For example, consider a record arriving with `recordedBy: Morten Hoefft | Tim Robertson`.  
It is not possible today, to search for `Tim Robertson` and discover this record, along with others having this value.

Terms that currently support auto-suggest are noted, which may bring additional considerations.

| **Term** | **Auto-suggest**                                      |
|----------|-------------------------------------------------------|
| datasetID  |  |
| datasetName  | yes (to be added) |
| otherCatalogNumbers  |  |
| protocol  |  |
| typeStatus  |  |
| recordedBy  | yes |
| identifiedBy  | yes |
| preparations  |  |
| samplingProtocol  | yes |
| locality  | yes |

This issue is intended to focus only on [existing fields in ES](https://github.com/gbif/pipelines/blob/master/gbif/pipelines/ingest-gbif-beam/src/main/resources/elasticsearch/es-occurrence-schema.json), and those already being added in work in progress, and not to propose additional fields.

MortenHofft:
is `locality` a multi value field? It isn't in https://dwc.tdwg.org/terms/#dwc:locality and it isn't obvious to me why it ought to be.

MortenHofft:
It believe it would be an appreciated feature if ordering could be retained (when serialized back into a string). 

Perhaps related: https://github.com/gbif/portal-feedback/issues/3292#issuecomment-1038830329 for the desire to keep ordering in the multimedia array.

sylmorin-gbif:
Hi,

I'm really wondering why datasetID and datasetName are multi-value?
(I just noticed this change in ALA since I'm migrating to the new ALA / Pipeline)

How can an occurrence be part of 2 datasets?

Thanks

timrobertson100:
> I'm really wondering why datasetID and datasetName are multi-value?

The request came from the GBIF node community, where (if I can remember correctly) they are using datasetID and datasetName to encode the various projects that a record is associated with, and also when dealing with aggregating data  into a single dataset for GBIF with multiple origins.

sylmorin-gbif:
Thanks for the explanation!

We need a multi-value projectID field ! (another similar topic about projectID in the metadata :-)

timrobertson100:
Yes. 

[...]

PR #669, opened 2022-02-15
#665 new fields ala

marcos-lg:
I did some changes in the ALA pipelines to adapt them to the changes in #665.

The changes that affect ALA are the ones done in the BasicRecord:
- TypeStatus changed from String to array
- New fields added (all of them arrays)

You can see the changes in the schema in this [commit](https://github.com/gbif/pipelines/commit/ddcf435a64c46d764ff22f86e7b8b178529e0461).

In this PR you can see the changes that I think they need to be done in ALA but you need to review them because I don't know if there are other implications.


djtfmartin:
LGTM

djtfmartin (approved):

Issue #652, opened 2022-01-27
Support running livingatlas pipelines on EMR


djtfmartin:
Support running livingatlas pipelines on Amazon's Elastic MapReduce platform.
This will require some changes to the way HDFS is referenced, in particular allowing the `hdfs:///` path to resolve to local HDFS.

PR #684, opened 2022-03-03
Work for #652  - livingatlas pipelines on EMR

djtfmartin:
Work for #652 - allowing `livingatlas` pipelines to be run on the AWS EMR environment.

The changes required were all to do with how HDFS paths are constructed.

EMR allows the use of the path `hdfs:///`, which allows access to HDFS without the DNS name of the master node  (which is different for each EMR cluster run).

These paths cause issues outside of Apache Beam, namely when there are usages of `org.apache.hadoop.fs.Path` which will throw a `IllegalArgumentException`  with the message "Wrong FS". This was noticed in a number of `livingatlas` pipelines which use HDFS outside of a beam pipeline,  in the writing of metrics files for example. Hence some changes where made to `FSUtils` in the `core` module. Changes where also made to `FileSystemFactory`  - so has potential impact to GBIF.

Also includes:
* minor change to `la-pipelines` script so that it can be used on non-debian environments (EMR = centos)
* config references to new GBIF vocabs
* removal of file locking for DWCA pipelines (no longer in use)
* configurable spark-submit location

This PR will be squashed and merged once reviewed.


muttcg (approved):
LTGM, I will run some extra tests and let you know in case of bugs from our side

Issue #685, opened 2022-03-04
Iivingatlas index pipeline - use hashing to avoid hotkeys in sampling join


djtfmartin:
The current `IndexRecordToSolrPipeline` uses a series of joins to pull together:

* IndexRecord
* Jackknife
* Clustering
* Expert distribution
* Sampling

Joining the sampling is problematic due to hotkeys. The join uses the latitude and longitude from the sampling to join to the index record which is problematic for some points which can have over 100k records in the worst case. To get around this, the IndexRecordToSolrPipeline currently uses partitions, but this is slow and with the addition of Expert distributions is proving to be unstable.

An alternative to partitions is to introduce a level of hashing in the join between IndexRecord and Sampling.

djtfmartin:
This work was merged into dev branch with #686

PR #686, opened 2022-03-09
685 sampling hashing

djtfmartin:
See #685 

This is a re-work of the SOLR indexing pipeline due to problems with running with the new distribution outliers.
This version does a full index of ALA on EMR in about 3 hours against the test SOLR cluster (4 SOLR cloud nodes - m2xg instances).
Previously this was taking 5hrs (with 4 SOLR cloud nodes - 3hrs in production with 8 nodes).

It avoids the partitioning of index record, and instead uses hashing of sampling records to avoid hot keys.
Includes an updated version of SOLRIO.java from beam with the addition of setting the ZK timeout.

### DIFF
--- a/livingatlas/pipelines/src/main/java/org/apache/beam/sdk/io/solr/SolrIO.java
+++ b/livingatlas/pipelines/src/main/java/org/apache/beam/sdk/io/solr/SolrIO.java
@@ -27,6 +27,7 @@ import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Optional;
 import java.util.function.Predicate;
 import org.apache.beam.sdk.annotations.Experimental;
 import org.apache.beam.sdk.annotations.Experimental.Kind;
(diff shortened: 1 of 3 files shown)


livingatlas/pipelines/src/main/java/org/apache/beam/sdk/io/solr/SolrIO.java
```diff
@@ -188,27 +187,15 @@ private void populateDisplayData(DisplayData.Builder builder) {
...
-      params.set(HttpClientUtil.PROP_BASIC_AUTH_USER, getUsername());
-      params.set(HttpClientUtil.PROP_BASIC_AUTH_PASS, getPassword());
-      return HttpClientUtil.createClient(params);
-    }
-
     AuthorizedSolrClient<CloudSolrClient> createClient() {
-      CloudSolrClient solrClient = new CloudSolrClient.Builder().withZkHost(getZkHost()).build();
+      CloudSolrClient solrClient =
+          new CloudSolrClient.Builder(ImmutableList.of(getZkHost()), Optional.empty()).build();
       solrClient.setZkClientTimeout(180000);
```
timrobertson100, line 193: @djtfmartin - is this the only addition, please?
Let's get this fixed in Beam

livingatlas/pipelines/src/main/java/org/apache/beam/sdk/io/solr/SolrIO.java
```diff
@@ -693,6 +682,14 @@ private void flushBatch() throws IOException, InterruptedException {
                     exception);
 
               } else {
+
+                LOG.warn("### Re-connecting to SOLR client in SOLRIO");
```
timrobertson100: @djtfmartin - is this also a SolrIO deviation?

livingatlas/pipelines/src/main/java/org/apache/beam/sdk/io/solr/SolrIO.java
djtfmartin: it is the only addition

livingatlas/pipelines/src/main/java/org/apache/beam/sdk/io/solr/SolrIO.java
djtfmartin: yep - ill remove

qifeng-bai (approved):

Issue #695, opened 2022-03-28
Re-interpretation of dataset after mapping to GRSciColl entry doesn't link occurrences to GRSciColl


ManonGros:
I just reinterpreted one of the datasets that I had mapped to a GRSciColl entry but the occurrences aren’t linked.

* Here is the GRSciColl entry: https://api.gbif.org/v1/grscicoll/collection/dbc640f7-0584-4a31-8e90-5eaea7cf0956
* The reinterpretation of the dataset: https://registry.gbif.org/dataset/b5cdf587-3342-48ec-9130-ba1281d7166f/ingestion-history
* What I would expect from the lookup tool: https://api.gbif.org/v1/grscicoll/lookup?institutionCode=MNHN&collectionCode=IK&datasetKey=b5cdf587-3342-48ec-9130-ba1281d7166f
* Example of what I get for occurrences: https://www.gbif.org/occurrence/3497363492

marcos-lg:
Fixed. all the caches are refreshed now

PR #697, opened 2022-03-30
set an expiry time for kv store caches

marcos-lg:
#695 

Not sure about the values we should put.

There is a `LocationFeatureKvStoreFactory` class that uses this cache too but I didn't change it because it seems nobody uses it.


muttcg (approved):
LGTM

PR #714, opened 2022-04-26
suggest to remove redundant dependencies already defined in dwca-io

jhpoelen:
dwca-io depends on dwc-api and gbif-common, so no need to declare all three as far as I can tell.

### DIFF
--- a/sdks/core/pom.xml
+++ b/sdks/core/pom.xml
@@ -115,19 +115,7 @@
     </dependency>
     <dependency>
       <groupId>org.gbif</groupId>
-      <artifactId>dwc-api</artifactId>
-    </dependency>
-    <dependency>
-      <groupId>org.gbif</groupId>
       <artifactId>gbif-parsers</artifactId>
-    </dependency>
-    <dependency>
-      <groupId>org.gbif</groupId>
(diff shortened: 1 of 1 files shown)


jhpoelen:
Interesting. 

As far as I can tell, 

dwca-io uses dwc-api and gbif-common. 

and sdks/core uses dwca-io . 

So, I am a bit confused as why dwc-api and gbif-common are declared in the sdks/core directly. Perhaps my source of confusion is that I tend to prune dependencies that are already declared transitively, even if they are not accessed through classes of that module that declares them.

timrobertson100:
As a general rule, we include direct dependencies so as to strongly declare versions. core is directly using classes in dwc-api, gbif-common and dwca-io which is why they're all listed with explicit versions.

The others listed in the post above look suspicious to me though (e.g. okio)

jhpoelen:
Ok, that makes sense. I can see how you'd choose to explicitly set the versions of your transitive dependencies in some property file. However, this does mean that you'd have to ensure that dwca-io dependency versions are in sync (or compatible with) with the ones you declare elsewhere. 

Ah ! The joys of dependency management ; )

PR #725, opened 2022-05-25
Added copy of drs to la-pipelines script

vjrj:
A step to download, copy and copy to dfs datasets, so people can start using pipelines in a LA portal more easily.

A sample of run `la-pipelines copy all --hdfs`:

image

I did also some minor changes in the script (duplicate license, and use of the logging utility instead of direct `echo`).

### DIFF
--- a/livingatlas/configs/la-pipelines.yaml
+++ b/livingatlas/configs/la-pipelines.yaml
@@ -112,6 +112,7 @@ general:
   hdfsSiteConfig: ""
   # Path to core-site-config.xml
   coreSiteConfig: ""
+  copyPath: "/data/dwca-export"
 
 # class: au.org.ala.pipelines.beam.ALADwcaToVerbatimPipeline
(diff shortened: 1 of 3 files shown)


djtfmartin (approved):

vjrj:
@djtfmartin it's just a very small contribution but it's some step we need to start using pipelines from the community ... thanks to all of you

timrobertson100:
Thanks Vicente. It’s great you’re looking at that.

vjrj:
Thanks Tim. For the record, this was done by [this part of biocache-store](https://github.com/AtlasOfLivingAustralia/biocache-store/blob/3bf0e8f5e3361ada5f20c8033edf50e8748eb41e/src/main/scala/au/org/ala/biocache/load/DwCALoader.scala#L114) and I think ALA is using a similar script but via Apache Airflow.

PR #734, opened 2022-06-15
Moved copyPath config in la-pipelines.yaml to a self section away from Pipelines general options

vjrj:
I introduced an issue with #725 adding the new configuration to the `general` section:

image

So I moved the new copy command configuration `copyPath` to a different location instead of the general pipelines section.


djtfmartin (approved):

djtfmartin:
LGTM

PR #736, opened 2022-06-16
732 Handle types and multi values for custom terms

sylmorin-gbif:
During the "solr" step, custom fields are correctly converted to the types specified in the SolR schema.
Multi values fields are split using | as separator.


djtfmartin (approved):
LGTM

Issue #738, opened 2022-06-16
Evaluate extending GBIF downloads to support Event downloads


fmendezh:
Since Event and Downloads rely on very similar data structure worth evaluating or proposing a pipelines extension support Event downloads based on current Occurrence pipelines

PR #739, opened 2022-06-16
Event hdfs downloads

fmendezh:
HDFS/Hive processing was generalised to produce occurrence and events records suitable for GBIF downloads
This depends on branches:

- [https://github.com/gbif/dwc-api/tree/event-hdfs-downloads](https://github.com/gbif/gbif-api/tree/event-hdfs-downloads)
- [https://github.com/gbif/dwc-api/tree/event-hdfs-downloads](https://github.com/gbif/dwc-api/tree/event-hdfs-downloads)
- [https://github.com/gbif/postal-service/tree/event-hdfs-downloads](https://github.com/gbif/postal-service/tree/event-hdfs-downloads)

https://github.com/gbif/pipelines/issues/738


sdks/models/src/main/avro/occurrence-hdfs-record.avsc
```diff
@@ -734,6 +734,26 @@
...
+  }, {
+    "name" : "datasettitle",
+    "type" : [ "string", "null" ]
+  }, {
+    "name" : "parenteventgbifid",
+    "type" : [ {
+      "type" : "array",
+      "items" : [ {
+        "type" : "record",
+        "name" : "ParentEventGbifId",
```
marcos-lg, line 746: shouldn't this be in lower case too? or no need?

gbif/coordinator/tasks/src/main/java/org/gbif/pipelines/tasks/occurrences/hdfs/OccurrenceHdfsViewCallback.java
```diff
@@ -0,0 +1,46 @@
...
+      MessagePublisher publisher,
+      CuratorFramework curator,
+      PipelinesHistoryClient historyClient,
+      ExecutorService executor) {
+    return new OccurrenceHdfsViewCallback(config, publisher, curator, historyClient, executor);
+  }
+
+  @Override
+  public String routingKey() {
+    return PipelinesHdfsViewBuiltMessage.ROUTING_KEY;
```
marcos-lg, line 44: I think this should be as it was before:

```
String routingKey =
        new PipelinesInterpretedMessage().setRunner(config.processRunner).getRoutingKey();
```

These clis are deployed with 2 different configs (one for standalone and the other for distributed) and they use different routing keys. It'd have to be the same for the `EventsOccurrenceHdfsViewCallback`

sdks/models/src/main/avro/occurrence-hdfs-record.avsc
fmendezh: this schema is generated from the occurrence project, I'll look to see how easy is change that to provide better names for record types

sdks/core/src/main/java/org/gbif/pipelines/core/io/AvroReader.java
```diff
@@ -162,6 +162,9 @@ private static <T extends Record> Map<String, T> readRecords(
   private static List<Path> parseWildcardPath(FileSystem fs, String path) {
     if (path.contains("*")) {
       Path pp = new Path(path).getParent();
+      if (!fs.exists(pp)) {
+        return Collections.emptyList();
+      }
```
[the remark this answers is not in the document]
fmendezh: throwing an exception broke the execution processing archives of events and occurrences and one of them were empty

PR #749, opened 2022-06-22
748 Use TaxonID for taxonomy interpretation

sylmorin-gbif:
Use TaxonID for taxonomy interpretation


vjrj:
We have DRs with wrong taxonIds, so maybe this should be configured optionally

Issue #759, opened 2022-07-07
EventDate in some ALA records,  like 01/01/2000 cannot be recognised  - Should be treated as dd/mm/yyy


qifeng-bai:
Muhmoud reported:
For eventDate with the format of dd/mm/yyyy or mm/dd/yyyy there are only 138,680 records that cannot be interpreted as the mm and dd parts are less than 13  and the algorithm cannot decide which one is month or day. In total, all good not many records to worry about and it process dd/mm/yyyy as expected when there is no ambiguity.

Issues identified:
If 'defaultDateFormat' is not defined in pipeline config,  we use the default ISO formats to parse date. 

`defaultDateFormat: DMYT,DMY`

Need to be defined for ALA pipelines

Also,  other code changes for ALA date quality check

PR #760, opened 2022-07-07
#759 parse ALA date with dd/mm/yyyy format

qifeng-bai:


livingatlas/pipelines/src/main/java/au/org/ala/pipelines/beam/ALAVerbatimToInterpretedPipeline.java
```diff
@@ -103,7 +101,8 @@ public class ALAVerbatimToInterpretedPipeline {
 
   public static void main(String[] args) throws IOException {
     VersionInfo.print();
-    String[] combinedArgs = new CombinedYamlConfiguration(args).toArgs("general", "interpret");
+    String[] combinedArgs =
```
djtfmartin, line 104: can remove this change ? looks like its only formatting. Suggest rolling it back

qifeng-bai: defaultDateFormat is added in

qifeng-bai:
> @qifeng-bai This test currently fails: https://github.com/gbif/pipelines/blob/ala_dmy_format/livingatlas/pipelines/src/test/java/au/org/ala/pipelines/interpreters/AlaTemporalInterpreterTest.java#L190

Hi, Dave.  It has been fixed now

PR #763, opened 2022-07-19
Allow custom external stateProvinces file

vjrj:
This PR allows custom external stateProvince file as [StateProvinceCentrePoints](https://github.com/gbif/pipelines/blob/dev/livingatlas/pipelines/src/main/java/au/org/ala/pipelines/vocabulary/StateProvinceCentrePoints.java) and [CountryCentrePoints](https://github.com/gbif/pipelines/blob/dev/livingatlas/pipelines/src/main/java/au/org/ala/pipelines/vocabulary/CountryCentrePoints.java).

Added a testing resource using the fictional https://en.wikipedia.org/wiki/New_Yorkistan


djtfmartin (commented):
LGTM

Issue #779, opened 2022-09-05
Interpret dwc:continent using a geospatial layer


MattBlissett:
The geocoder now has a continent layer, see https://api.gbif-uat.org/v1/geocode/debug/map.html, which is a non-political 7-continent model, containing only land.  (It will later complement a sea/ocean layer, see #612).

![continents-bitmap-cache](https://raw.githubusercontent.com/gbif/geocode/master/geocode-ws/src/main/resources/org/gbif/geocode/ws/layers/continent.png)

Interpretation is done *after* country/coordinate interpretation, so if that flips coordinates or similar we use the new coordinates. For non-georeferenced occurrences, the existing continent parser is used, followed by a check that the continent is appropriate for the given country -- only Europe for Switzerland, Europe or Asia for Turkiye, Russia, Kazakhstan etc.

Note this means non-georeferenced marine occurrences will incorrectly have a continent assigned.

Main issue: https://github.com/gbif/parsers/issues/26

Would also resolve https://github.com/gbif/portal-feedback/issues/3642, probably https://github.com/gbif/portal-feedback/issues/2276, https://github.com/gbif/portal-feedback/issues/3580, https://github.com/gbif/portal-feedback/issues/3620, https://github.com/gbif/portal-feedback/issues/3526, https://github.com/gbif/portal-feedback/issues/2387, https://github.com/gbif/portal-feedback/issues/3620

Related also to https://github.com/gbif/portal-feedback/issues/3324

MattBlissett:
Completed and in prod.

PR #780, opened 2022-09-05
#779 Interpret dwc:continent

MattBlissett:
This interprets the dwc:continent, see #779, but it breaks an ALA test -- I'm not sure what you'll want to do, the existing continent interpretation wasn't so good. (Is ALA using GBIF's geocoder? In Docker? If not I can send the continent layer I made using [continents](https://github.com/gbif/continents/) in any suitable format if that's helpful.)


sdks/core/src/main/java/org/gbif/pipelines/core/parsers/location/parser/ContinentParser.java
```diff
@@ -0,0 +1,148 @@
...
+      ParsedField<Continent> field, Set<String> issues) {
+    if (!field.isSuccessful()) {
+      issues.addAll(field.getIssues());
+    }
+
+    return Optional.ofNullable(field.getResult());
+  }
+
+  public static void accept(Location l) {
+    if (l.getType() != null && l.getDistance() != null && l.getDistance() == 0) {}
```
muttcg, line 146: Does this supposed to be empty? Looks strange

MattBlissett: Oops, copy-paste error. The method isn't used, I'll remove it.

muttcg (approved):

PR #787, opened 2022-09-20
la-pipelines copy improvements

vjrj:
This PR improve the `la-pipelines` script:

- Fix the `copy --exclude="dr1,dr2"`  that was only skipping the first dr.
- Check that the downloaded dr  is a valid zip. If not it tries to clean it and its directory so other la-pipelines steps does not fails because of this empty dr.  

Skipping two drs and an incorrect zip:
 
image

and:
 
image


djtfmartin:
thanks @vjrj

PR #792, opened 2022-09-28
Fix for missing --master arg in la-pipelines. Remove duplicate code

vjrj:
Fix missing `--master` argument as discussed in slack. Remove duplicate code so all commands use the same spark-submit function. 

@djtfmartin I'm testing right now, if you want to wait a bit to merge.


vjrj:
Tested and works as expected.

djtfmartin:
thanks @vjrj

Issue #797, opened 2022-10-06
Change livingatlas integration tests to use testcontainers


djtfmartin:
Adding this for tracking - WIP.

The `livingatlas` module should use testcontainers instead of fabric8 maven plugin for docker based integration tests.
This would bring it inline with ITs in other pipeline modules e.g. `validator`

relates to #790

djtfmartin:
This was completed in #797

PR #799, opened 2022-10-07
Use testcontainers for `livingatlas` module

djtfmartin:
See #797


charvolant (approved):
Well, that was a marathon.
I assume comments in JSON is OK.

PR #824, opened 2022-11-18
Fix for yq version detection 

sadeghim:
la-pipelines was failing due to `unsupported yq version` error which was caused by version change in yq. It now includes a `v` as a prefix.
This commit is to fix this issue.

### DIFF
--- a/livingatlas/scripts/la-pipelines
+++ b/livingatlas/scripts/la-pipelines
@@ -22,7 +22,7 @@ fi
 
 # https://mikefarah.gitbook.io/yq/upgrading-from-v3
 YQ_VERSION="$(yq -V  | sed 's/.* //g')"
-if [[ $YQ_VERSION =~ ^3 ]]
+if [[ $YQ_VERSION =~ ^[a-zA-Z]*3\.[0-9]*\.[0-9]* ]]
 then
     function YQ_VERIFY() {
         cat "$1" | yq v -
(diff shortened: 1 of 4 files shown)


sdks/core/src/main/java/org/gbif/pipelines/core/interpreters/extension/MultimediaInterpreter.java
```diff
@@ -160,6 +160,8 @@ private static void parseAndSetType(Multimedia m, String v) {
       m.setType(MediaType.Sound.name());
     } else if (prFn.test("image", MediaType.StillImage)) {
       m.setType(MediaType.StillImage.name());
```
djtfmartin, line 162: are these changes intended to be included in this PR ?
I think we need to reference a separate issue.

djtfmartin (changes requested):
I suspect there are changes here that arent intended to be part of this PR.

sdks/core/src/main/java/org/gbif/pipelines/core/interpreters/extension/MultimediaInterpreter.java
muttcg: I made multimedia changes, so they mustn't be part of PR

livingatlas/scripts/la-pipelines
```diff
@@ -22,7 +22,7 @@ fi
 
 # https://mikefarah.gitbook.io/yq/upgrading-from-v3
 YQ_VERSION="$(yq -V  | sed 's/.* //g')"
-if [[ $YQ_VERSION =~ ^3 ]]
+if [[ $YQ_VERSION =~ ^[a-zA-Z]*3\.[0-9]*\.[0-9]* ]]
```
vjrj, line 25: As this is something that started in some point of version 4 and adding a `v`, maybe we can let this part untouched (as was working with version 3).

djtfmartin:
thanks for this @sadeghim 
This [PR](https://github.com/gbif/pipelines/pull/832) has now be merged which includes the single change that we need without the additional changes.

PR #865, opened 2023-03-08
LivingAtlas: Additional fields for `SpeciesListPipeline` (ARGA)

nickdos:
Pls assign to @djtfmartin.

PR contains changes for additional fields to be processed during the `specieslist` phase of the `index` pipeline. There are two set of fields that relate to:
- `locatedInCountry` a string field contains a single value, the (ISO) country name, populated via a species list. The idea being that the taxon is known to be located in that country. ARGA uses this for the large percentage of records that have no location data. Config var `includePresentInCountry` has a default `false` value.
- Traits from [AusTraits](https://austraits.org/) that are multi-value type and are also populated via species lists - one list per trait. Lists must have type `COMMON_TRAIT` and contain additional columns: `traitName` and `traitValue`. Traits defined in `IndexedFields.java` and (SOLR) `schema.xml` will be indexed as multi-value but any additional lists (that are added without changes to those files) will still be indexed as dynamic fields. Config var `includeTraits` has a default `false` value.


livingatlas/pipelines/src/main/java/au/org/ala/pipelines/util/SpeciesListUtils.java
```diff
@@ -1,9 +1,7 @@
 package au.org.ala.pipelines.util;
 
 import com.google.common.base.Strings;
-import java.util.ArrayList;
-import java.util.Iterator;
-import java.util.List;
+import java.util.*;
```
nickdos, line 4: Just noticed this - IDE did this and it might break coding rules?

livingatlas/pipelines/src/main/java/au/org/ala/pipelines/transforms/IndexFields.java
```diff
@@ -73,4 +74,7 @@ public interface IndexFields {
   String GGBN_TERMS_LOAN = "http://data.ggbn.org/schemas/ggbn/terms/Loan";
   String LOAN_DESTINATION_TERM = "http://data.ggbn.org/schemas/ggbn/terms/loanDestination";
   String LOAN_IDENTIFIER_TERM = "http://data.ggbn.org/schemas/ggbn/terms/loanIdentifier";
+  String AUS_TRAITS_FIRE_RESPONSE = "fire_response";
```
djtfmartin, line 77: Convert from snake case to camel

livingatlas/pipelines/src/main/java/au/org/ala/pipelines/transforms/IndexFields.java
```diff
@@ -51,6 +51,7 @@ public interface IndexFields {
   String POINT_0_02 = "point-0.02";
   String POINT_0_1 = "point-0.1";
   String POINT_1 = "point-1";
+  String PRESENT_IN_COUNTRY = "presentInCountry";
```
djtfmartin, line 54: Wondering why we need this. Cant the data just provide countryCode ?

nickdos: Its the field name, so data will look like `presentInCountry:Australia` or `presentInCountry:Italy`. Could use country code I suppose `presentInCountry:AU` but data is from `region` field in species list, which uses full name, so would require an additional lookup.

nickdos: I've renamed the field to `taxonPresentInCountry` now.

Issue #205, opened 2019-07-27
Consider using an identification from extension when missing from core


timrobertson100:
Pipelines should mirror the outcome of https://github.com/gbif/occurrence/issues/126

This may or may not require coding changes in this project as well, depending on the outcome.

marcos-lg:
Deployed to PROD.

PR #866, opened 2023-03-10
#205 taking values from identification extension when missing from core

marcos-lg:
#205 Basically, if it's an identification field and all the identification fields in the core are null we take the values of these fields from the Identification extension if exists. If there are multiple identification extensions we take it from the most recent one but only if we can determine which one it's the most recent.

### DIFF
--- a/sdks/core/src/main/java/org/gbif/pipelines/core/parsers/temporal/TemporalParser.java
+++ b/sdks/core/src/main/java/org/gbif/pipelines/core/parsers/temporal/TemporalParser.java
@@ -209,4 +209,34 @@ public class TemporalParser implements Serializable {
 
     return likelyRange.contains(LocalDate.of(year, month, day));
   }
+
+  public static Optional<Long> getTime(TemporalAccessor temporalAccessor) {
+    if (temporalAccessor == null) {
+      return Optional.empty();
+    }
+
+    int year;
+    if (temporalAccessor.isSupported(ChronoField.YEAR)) {
+      year = temporalAccessor.get(ChronoField.YEAR);
+    } else {
+      return Optional.empty();
+    }
+
+    int month = 1;
+    if (temporalAccessor.isSupported(ChronoField.MONTH_OF_YEAR)) {
+      month = temporalAccessor.get(ChronoField.MONTH_OF_YEAR);
+    }
+
+    int day = 1;
+    if (temporalAccessor.isSupported(ChronoField.DAY_OF_MONTH)) {
+      day = temporalAccessor.get(ChronoField.DAY_OF_MONTH);
+    }
+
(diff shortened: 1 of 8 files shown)


sdks/core/src/main/java/org/gbif/pipelines/core/parsers/temporal/TemporalParser.java
```diff
@@ -209,4 +209,34 @@ protected static boolean isValidDate(
...
     return likelyRange.contains(LocalDate.of(year, month, day));
   }
+
+  public static Optional<Long> getTime(TemporalAccessor temporalAccessor) {
+    if (temporalAccessor == null) {
+      return Optional.empty();
+    }
+
+    int year;
+    if (temporalAccessor.isSupported(ChronoField.YEAR)) {
```
timrobertson100, line 219: Are these changes intended to be in this commit?

timrobertson100:
The logic in here looks good to me. 
I think a few methods intended for chrono work have slipped into the commit by mistake which might be worth checking.

sdks/core/src/main/java/org/gbif/pipelines/core/parsers/temporal/TemporalParser.java
timrobertson100: OK - it was just the reference to `ChronoField` that jumped out as chrono isn't used in the [identification history extension](https://rs.gbif.org/extension/dwc/identification_history_2022-02-02.xml). Made me just wonder if that was unrelated work slipping in

marcos-lg: Ah it's because I use the `dateIdentified ` field of the extension for the sorting and since it's a string I have to parse it and convert it into a date.

sdks/core/src/main/java/org/gbif/pipelines/core/utils/ModelUtils.java
```diff
@@ -106,4 +148,70 @@ public static void addIssue(Issues model, Set<String> issues) {
   public static void addIssueSet(Issues model, Set<OccurrenceIssue> issues) {
     issues.forEach(x -> addIssue(model, x));
   }
+
+  public static Map<String, String> getIdentificationFieldTermsSource(ExtendedRecord er) {
```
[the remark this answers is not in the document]
marcos-lg: ok, I'll move it to a class in the same package and I'll also move the function in the temporalParser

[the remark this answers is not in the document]
marcos-lg: just pushed it

muttcg (approved):

Issue #888, opened 2023-05-03
Record-level bulk annotations from publications


djtfmartin:
This an issue to track experimental work on linking specific publications that have reused ALA data and republished cleaned version of the data. The work would be to create a pipeline that links that data using UUIDs from the republished records.

djtfmartin:
Released in version 2.15.1

PR #890, opened 2023-05-03
Work for #888 - record annotations

djtfmartin:
This PR contains a new pipeline `AnnotationsPipeline` and changes to SOLR indexing to include annotations as child documents.


patkyn (approved):
Looks good

Issue #862, opened 2023-03-02
Support a seedbank extension


djtfmartin:
ALA is working on supporting for seedbank data. Currently this is achieved through using Event core archives. MeasurementsOrFacts are used to store a variety of properties associated with Seedbank processes (Accessions, Trails, Treatment).
An arguably better more descriptive method would be to define a draft Darwin core extension and add transforms + support for this in pipelines (initially only for use in the `livingatlas` module).

PR #892, opened 2023-05-10
Seedbank extensions support #862

djtfmartin:
Work seedbank extension support. Includes:
* Basic parsing of seedbank fields
* Elastic search indexing


livingatlas/pipelines/src/main/java/au/org/ala/pipelines/interpreters/SeedbankInterpreter.java
```diff
@@ -0,0 +1,289 @@
...
+    Objects.requireNonNull(er);
+    Objects.requireNonNull(sr);
+    ExtensionInterpretation.Result<SeedbankRecord> result = HANDLER.convert(er);
+  }
+
+  public static void setSampleWeight(SeedbankRecord sr, String value) {
+    try {
+      sr.setSampleWeightInGrams(Double.parseDouble(value));
+    } catch (Exception e) {
+      // do nothing
```
charvolant, line 75: Should this and the other instances be adding an issue somewhere, since it's unexpected data?

charvolant (approved):
Approved. A query about swallowing non-doubles during interpretation.

livingatlas/pipelines/src/main/java/au/org/ala/pipelines/interpreters/SeedbankInterpreter.java
djtfmartin: Yes, i think you are right. I'll create an issue for that

Issue #893, opened 2023-05-10
Events - provide addition indexing of Surveys


djtfmartin:
It is useful to treat the eventType='Survey' as a special type of event to report against.
This issue is for tracking changes to elastic indexing to allow breakdowns against surveys.

djtfmartin:
Released in 2.15.1

PR #894, opened 2023-05-10
SurveyID indexing - Work for #893

djtfmartin:
This PR just adds indexing of `surveyID` as a separate field.
Also includes:
* a change to import the correct version of FileUtils
* adding the ALAMetadataTransform to the java only interpretation pipeline


charvolant (approved):
All looks good.

Issue #882, opened 2023-04-20
Have GitHub issues instead of the "Pipelines identifier workflow failed" emails


ManonGros:
Hi @muttcg,

@CecSve and I are having a hard time keeping track of all the datasets blocked in the pipeline because of occurrenceID change. They are many of them and we receive new emails every new ingestion.
It is hard to know which publisher has already been contacted and by whom (Cecilie John or I).

Instead of the email, could we have a GitHub repository?
The idea would be to have a new issue for every "new" failing dataset and only add a tag for the failing datasets that have already an open issue, like in this repo: https://github.com/gbif/collections-sync/issues
Would that be possible?

Tagging @ahahn-gbif for FYI

ManonGros:
We would like:
1. A GitHub repository dedicated to these issues
2. Have one issue per datasetKey and labels added every time there is new problems detected on the same dataset
3. If we close an issue and new problems arise, we want a new issue
4. The issues should contain the same text as the email notification (see example below) + if possible:
    * the name and/or link/UUID of the publisher
    * the name and/or UUID and/or link to the installation (so if several datasets from the same installation are failing, we can contact the installation directly in one go)
    * A label for the endorsing node or publishing country

Example of email notification we currently get:
> Identifier validation failed for the dataset Artrópodos de los Parques Distritales Ecológicos de Montaña, crawler attempt: 61 cause:
>
> GBIF IDs hit maximum allowed - 50%, duplicates - 100%, total records count 1453, absent records count 1453
>
> You can skip/fix identifier validation using [Registry UI](https://registry.gbif.org/dataset/865c11e6-8d75-4acb-8f09-5d909e08ed6b/ingestion-history)

timrobertson100:
@ManonGros - should the repository be open or private, please?
Any suggestions on the name? How about `gbif/ingestion-management` ?

ManonGros:
The repo should be open so we can send the issues to Node Managers. `gbif/ingestion-management` sounds good to me.

timrobertson100:
https://github.com/gbif/ingestion-management/issues is set up, and @ManonGros is an admin. I've disabled wiki, and actions but we should consider what we can do in e.g. the README and the issue template to stop people using it as a general dumping ground for helpdesk questions.

marcos-lg:
It's deployed to PROD although we test it with one dataset and noticed that the step metrics are empty even though they exist. I created an issue in the registry to track it: https://github.com/gbif/registry/issues/505

PR #918, opened 2023-06-30
changed registry ws call to notify about identifiers validation issues

marcos-lg:
Changes for #882 

The changes in the registry are deployed to DEV and UAT and will be deployed to prod at the beginning of the next week.

This branch `#882_gh_issues` was created from the [pipelines-parent-2.14.5 tag](https://github.com/gbif/pipelines/releases/tag/pipelines-parent-2.14.5) so we can use it to do the release.


marcos-lg:
The changes in the registry are already deployed to PROD.

muttcg (approved):

Issue #921, opened 2023-07-04
Improve species lookup


timrobertson100:
Recording these here but changes in KVS will be needed to address it.

- [x] Stop interpreting rank before calling the species lookup service (see [justification](https://github.com/gbif/checklistbank/issues/270)) in the production and pre-backbone check pipelines unless it is specifically given in the `rank` field
- [x] investigate if we have key collisions for e.g. `name=foobar` and `genus=foo&species=bar` in the cache because of [this](https://github.com/gbif/key-value-store/blob/dev/kvs-gbif/src/main/java/org/gbif/kvs/species/SpeciesMatchRequest.java#L57).

PR #923, opened 2023-07-06
#921 Stop interpreting rank and incertae sedis instead of null kingdoms

timrobertson100:
This brings in the KV store change you just reviewd @mdoering which will go into indexing and the pre-backbone release script. I also fixed up the null kingdom to incertae sedis to reduce the amount of noise


mdoering (approved):
All good, thanks!

timrobertson100:
Thanks @mdoering

Issue #217, opened 2018-07-18
DwC field scientificNameID is not used at all


bart-v:
**This issue was moved from portal-feedback to pipelines**

Example
https://www.gbif-uat.org/occurrence/search?dataset_key=740cf4e0-37ca-4389-ba8f-4e1bc5177893&taxon_key=5401803

Lists the records as Oligochaeta and appends the authority "K.Koch" just like that.
That makes these marine occurrences terrestrial plants... 

While a scientificNameID urn:lsid:marinespecies.org:taxname:2036 is provided, that can be resolved to the animal class Oligochaeta.

This is a missed chance to fix homonyms in an easy way...

timrobertson100:
For current links, all [Danish Mycological Society, fungal records database](https://www.gbif.org/dataset/84d26682-f762-11e1-a439-00145eb45e9a) records contain `scientificNameID` pointing to [Index Fungorum](http://www.indexfungorum.org/Names/NamesRecord.asp?RecordID=327981) such as [this example](https://www.gbif.org/occurrence/2237852364).

bart-v:
@mdoering about finding out what checklist (version) has been used, everything is solved by using a proper and persistent GUID (like LSID): it tells you what authority has been used, on a per record basis.

I don't understand this question
> if we rely on globally unique ids (...) how do we know which checklist is the authority in case several checklists use these ids?

If it's a GUID, there is only one single checklist who has assigned/generated this GUID, so there is nothing to choose from?

bart-v:
Thanks @timrobertson100

mdoering:
@bart-v a properly versioned LSID would tell you what it was when resolving it. But I doubt a DwC WoRMS archive contains all historical versions of a name or deleted names. 

My point about a non unique GUID is that there might be various datasets, e.g. molluscabase, WoRMS, Catalogue Of Life that all use the same GUID. Knowing which is the authorative one seems trivial by looking at the domain, but I would expect we better have some metadata about that on the dataset level. I am sure GUIDs will not appear once only.

bart-v:
WoRMS could do versions but that is usually overkill. We hardly ever change names, but create new ones ans point to them to each other. We do keep track of deletions.

I agree that some metadata on dataset level is needed, indeed.

auspex:
[...]

PR #943, opened 2023-09-06
#217 use scientificNameID, taxonID and taxonConceptID

timrobertson100:
All the work is done in the KVS project so this should really add configuration, use the 3 *ID terms from occurrence and then use the new class name for the API call .


timrobertson100:
This is missing Issues I see. Closing

Issue #978, opened 2023-10-31
Refactor clustering in Java


timrobertson100:
There is no need for clustering to use scala. Refactor this into Java for consistency.

PR #988, opened 2023-11-21
#978 Rewrite of clustering driver in Java

timrobertson100:


timrobertson100:
Discussed with Nik and OK to self-merge. Does not affect ALA.

Issue #989, opened 2023-11-21
Clustering: automate in an Oozie workflow


timrobertson100:
Automate the running of clustering in a weekly (configurable) Oozie job

timrobertson100:
In production

PR #990, opened 2023-11-22
#989 Oozie workflow for clustering

timrobertson100:
This provides an Oozie workflow to run clustering and adds the steps to truncate the HBase table and load the HFiles in place.


timrobertson100:
Self merging after discussion with Nik

Issue #729, opened 2022-05-30
Re-add the ability to load DwCAs from HDFS


djtfmartin:
During the EMR work, the ability to run the DWCA -> AVRO pipeline against archives stored in HDFS was dropped as this is not used in EMR deployments. However, it is still required in non-EMR deployments.
This functionality needs to be added back in to the DWCA pipeline

djtfmartin:
merged into `dev`

vjrj:
Thanks @djtfmartin for this.

vjrj:
This started to fail in https://github.com/gbif/pipelines/pull/809 where:

![image](https://github.com/gbif/pipelines/assets/180085/04f82f47-090d-48de-a5e2-c4683ce4e117)

in `livingatlas/pipelines/pom.xml`. I was a bit lost because` hadoop-core` is in parent `pom.xml`.

The last hdfs working build is `2.13.0-SNAPSHOT+0~20221123105804.867~1.gbp85d7b3`.

PR #1004, opened 2023-12-20
Interpret fix for hdfs support

vjrj:
This fix the #729 that I reopened (I though it was a regression), see https://github.com/gbif/pipelines/issues/729#issuecomment-1850110410.


djtfmartin (commented):
LGTM

PR #1026, opened 2024-01-24
updated versions for camtrap

marcos-lg:
We need to update versions to support the new endpoint type of `CAMTRAP_DP`. It's just a new value in the enum(actually it was renamed).


muttcg (approved):
lgtm

PR #1080, opened 2024-07-24
geoTime vocabulary

marcos-lg:
Needs this branch of gbif-api to compile https://github.com/gbif/gbif-api/tree/geotime_issues

Can't be merged until the vocab is ready.


sdks/core/src/main/java/org/gbif/pipelines/core/interpreters/core/GeologicalContextInterpreter.java
```diff
@@ -150,25 +165,201 @@ private static void interpretTerm(
...
+        .filter(v -> v.length == 2)
+        .filter(
+            v ->
+                Arrays.asList(RANK_TAG, START_AGE_TAG, END_AGE_TAG)
+                    .contains(v[0].toLowerCase().trim()))
+        .collect(Collectors.toMap(v -> v[0].toLowerCase().trim(), v -> v[1].toLowerCase().trim()));
+  }
+
+  private static TermData getTermData(DwcTerm term) {
+    if (term.equals(DwcTerm.earliestEonOrLowestEonothem)
```
muttcg, line 290: I think you don't need equal for Enums.
Maybe use switch?
```
private static TermData getTermData(DwcTerm term) {
    String rank = null;
    OccurrenceIssue inferedFromParent = null;
    OccurrenceIssue rankMismatch = null;
    OccurrenceIssue invalidRange = null;
    OccurrenceIssue parentMismatch = null;
    switch (term) {
      case earliestEonOrLowestEonothem:
      case latestEonOrHighestEonothem:
        rank = EON;
        rankMismatch = OccurrenceIssue.EON_OR_EONOTHEM_RANK_MISMATCH;
        invalidRange = OccurrenceIssue.EON_OR_EONOTHEM_INVALID_RANGE;
        break;
      case earliestEraOrLowestErathem:
      case latestEraOrHighestErathem:
        rank = ERA;
        inferedFromParent = OccurrenceIssue.ERA_OR_ERATHEM_INFERRED_FROM_PARENT_RANK;
        rankMismatch = OccurrenceIssue.ERA_OR_ERATHEM_RANK_MISMATCH;
        invalidRange = OccurrenceIssue.ERA_OR_ERATHEM_INVALID_RANGE;
        parentMismatch = OccurrenceIssue.EON_OR_EONOTHEM_AND_ERA_OR_ERATHEM_MISMATCH;
        break;
      case earliestPeriodOrLowestSystem:
      case latestPeriodOrHighestSystem:
        rank = PERIOD;
        inferedFromParent = OccurrenceIssue.PERIOD_OR_SYSTEM_INFERRED_FROM_PARENT_RANK;
        rankMismatch = OccurrenceIssue.PERIOD_OR_SYSTEM_RANK_MISMATCH;
        invalidRange = OccurrenceIssue.PERIOD_OR_SYSTEM_INVALID_RANGE;
        parentMismatch = OccurrenceIssue.ERA_OR_ERATHEM_AND_PERIOD_OR_SYSTEM_MISMATCH;
        break;
      case earliestEpochOrLowestSeries:
      case latestEpochOrHighestSeries:
        rank = EPOCH;
        inferedFromParent = OccurrenceIssue.EPOCH_OR_SERIES_INFERRED_FROM_PARENT_RANK;
        rankMismatch = OccurrenceIssue.EPOCH_OR_SERIES_RANK_MISMATCH;
        invalidRange = OccurrenceIssue.EPOCH_OR_SERIES_INVALID_RANGE;
        parentMismatch = OccurrenceIssue.PERIOD_OR_SYSTEM_AND_EPOCH_OR_SERIES_MISMATCH;
        break;
      case earliestAgeOrLowestStage:
      case latestAgeOrHighestStage:
        rank = AGE;
        inferedFromParent = OccurrenceIssue.AGE_OR_STAGE_INFERRED_FROM_PARENT_RANK;
        rankMismatch = OccurrenceIssue.AGE_OR_STAGE_RANK_MISMATCH;
        invalidRange = OccurrenceIssue.AGE_OR_STAGE_INVALID_RANGE;
        parentMismatch = OccurrenceIssue.EPOCH_OR_SERIES_AND_AGE_OR_STAGE_MISMATCH;
        break;
      default:
    }
    return TermData.of(rank, inferedFromParent, rankMismatch, invalidRange, parentMismatch);
  }
  ```

File: sdks/core/src/main/java/org/gbif/pipelines/core/converters/OccurrenceJsonConverter.java [hunk 2]
```diff
@@ -200,10 +204,42 @@ private void mapBasicRecord(OccurrenceJsonRecord.Builder builder) {
       JsonConverter.convertVocabularyConcept(gx.getLatestAgeOrHighestStage())
           .ifPresent(gcb::setLatestAgeOrHighestStage);
 
+      GeologicalRange.Builder rangeBuilder = GeologicalRange.newBuilder();
```
muttcg, line 207: Can it be the case when START_AGE_TAG and START_AGE_TAG have different types? Fx:
START_AGE_TAG - getEarliestAgeOrLowestStage
END_AGE_TAG - getLatestPeriodOrHighestSystem

sdks/core/src/main/java/org/gbif/pipelines/core/converters/OccurrenceJsonConverter.java
```diff
@@ -200,10 +204,42 @@ private void mapBasicRecord(OccurrenceJsonRecord.Builder builder) {
...
+              gx.getLatestEraOrHighestErathem(),
+              gx.getLatestEonOrHighestEonothem())
+          .ifPresent(rangeBuilder::setGt);
+      gcb.setRange(rangeBuilder.build());
+
       builder.setGeologicalContext(gcb.build());
     }
   }
 
+  private Optional<String> chooseFirst(String tag, VocabularyConcept... vocabularyConcepts) {
```
muttcg, line 230: It looks a bit like interpretation part, so maybe add 2 fileds to GeologicalContext avro and fill range in GeologicalContextInterpreter

sdks/core/src/main/java/org/gbif/pipelines/core/config/model/VocabularyConfig.java
```diff
@@ -18,14 +18,13 @@ public class VocabularyConfig implements Serializable {
   // directory where the vocabulary files are stored
   private String vocabulariesPath;
 
-  private Map<Term, String> vocabulariesNames = Collections.emptyMap();
+  private Map<String, String> vocabulariesNames = Collections.emptyMap();
```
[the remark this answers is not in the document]
marcos-lg: This was because this vocabulary is used for 10 terms so instead of adding the 10 terms to the map I thought to use this as key https://github.com/gbif/dwc-api/blob/c366f91c5cb0cb2eb68d40b9567fe45b9a9ead0a/src/main/java/org/gbif/dwc/terms/DwcTerm.java#L279C30-L279C53 

And then do this check: https://github.com/gbif/pipelines/blob/41ed816fbb3516436805e62472f455df86d56bf7/sdks/core/src/main/java/org/gbif/pipelines/core/parsers/vocabulary/VocabularyService.java#L23

What do you think?

sdks/core/src/main/java/org/gbif/pipelines/core/converters/OccurrenceJsonConverter.java
marcos-lg: I actually did it on purpose so if, for example, the earliest is provided but not the latest I take the latest from the parent but now I'm realising that it makes more sense to just take both start and end from the term provided(in case there is only one), the earliest in this example.

So I'll change it

sdks/core/src/main/java/org/gbif/pipelines/core/converters/OccurrenceJsonConverter.java
marcos-lg: makes sense

sdks/core/src/main/java/org/gbif/pipelines/core/utils/VocabularyUtils.java
```diff
@@ -0,0 +1,57 @@
...
+  public static VocabularyConcept getConcept(LookupConcept c, Map<String, String> tagsMap) {
+    return getConcept(c.getConcept().getName(), c.getParents(), tagsMap);
+  }
+
+  public static VocabularyConcept getConcept(
+      String conceptName, List<LookupConcept.Parent> parents, Map<String, String> tagsMap) {
+    // we sort the parents starting from the top as in taxonomy
+    List<String> sortedParents =
+        parents.stream().map(LookupConcept.Parent::getName).collect(Collectors.toList());
+    Collections.reverse(sortedParents);
```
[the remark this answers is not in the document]
marcos-lg: Actually if I do this it sorts the parents alphabetallicay inversed but we need to keep the order that we get from the vocabulary lookup. It's how it's working now in prod

muttcg (approved):
Thanks for the changes. I think rest of all is ok

muttcg (approved):

PR #1159, opened 2025-05-14
Living Atlas Multimedia Extension issue

StefanVanDyck:
Fixes issue processing iNaturalist GBIF exports in the living atlas pipelines.

The multimedia extension does not have an outright occurrenceId column, so the existing logic was mapping the extension to a null value.


timrobertson100 (approved):
This looks a reasonable thing to do, to me

djtfmartin (approved):

Issue #1048, opened 2024-03-19
K8s: Getting rid of Apache Beam


muttcg:
Apache Beam appears to be a redundant extra abstraction layer for pipelines, not providing significant benefits. Moreover, the combination of the latest Beam version with the pipelines-k8s version introduces bugs, complicating the diagnosis of actual issues.

Avro: https://spark.apache.org/docs/latest/sql-data-sources-avro.html#load-and-save-functions
Elasticseach: https://www.elastic.co/guide/en/elasticsearch/hadoop/current/spark.html

timrobertson100:
I've done some exploration(!) into this [here](https://github.com/timrobertson100/pipelines2). That project is currently able to run the Basic and Taxonomy transformations from Avro into the specific objects on both Spark Connect and using `spark-submit`.

Some learnings / notes about the current state:

1. Spark Connect has limitations, but is fabulous to develop against with a locally running server.
2. Pipeline configuration is all built on Beam configuration. I've used a basic Jackson yaml parser instead. This seems a promising approach.
3. I initially went for simple SQL UDFs [here](https://github.com/timrobertson100/pipelines2/releases/tag/v0.1). However, we have fairly complex code for e.g. extracting the identification, which won't port well (at all?) to simple SQL.
4. Spark does not natively provide SerDe for the Specific Avro objects. I ported those all to POJOs so the spark BeanEncoder can be used. Note that a lot of things are removed in Spark Connect (e.g. Kryo serialization).
5. I wanted to try and remove the HBase cache dependency, and instead use a "distinct, lookup, join" approach. This seems to work reasonably well so far
6. I have not ported any tests (yet)

PR #1175, opened 2025-06-06
Spark only processing

timrobertson100:
Note this targets the `spark` branch, and not dev.

This brings in:
1. A Plugin that generates POJOs for all Avro schemas
2. A example that uses Spark only (not Beam) to demonstrate how they are used for basic ingestion

I tried without success to use the Avro objects in Spark Connect (including 4.0.0) due to the lacking spark Avro `Encoder`. Our entire interpretation is built around specific Avro objects. This PR gives the ability to replace those Avro objects with equivalent Beans, with the minimal amount of changes needed.

There would still be much work needed to complete this, including likely removal of ALA code into a separate place.

For now, I suggest to merge this into the Spark branch, and try it for yourselves noting what happens in the model-pojos and the spark-ingest.

Relates to #1048.


timrobertson100:
> LGTM - maybe remove Transform classes & move logic to Interpreters

Yeah, I think there is a lot to clean up. I was trying to do the minimum changes so it could be maintained alongside dev, and then be something to then progressively clean up - if we were to commit to it. If everything gets refactored on this branch, we'd never merge it.

I didn't port over the taxonomy stuff because it has all changed since I last did it. If you care to, please feel free to port this over https://github.com/timrobertson100/pipelines2/blob/main/interpretation-spark/src/main/java/org/gbif/pipelines/interpretation/spark/TaxonomyInterpretation.java making the changes needed for the new lookup service.

PR #1226, opened 2025-11-18
Fix reading json dynamicproperties field when escaped by CSV writer.

StefanVanDyck:
For example when exporting from GBIF.
I also looked into changing te CSV parsing to cleanup the values for us.
But this logic is deep-down inside gbif-common / gbif-dwca-io and I suspect we might break a lot more than we would actually fix.

### DIFF
--- a/livingatlas/pipelines/src/test/java/au/org/ala/pipelines/transforms/IndexRecordTransformTest.java
+++ b/livingatlas/pipelines/src/test/java/au/org/ala/pipelines/transforms/IndexRecordTransformTest.java
@@ -1,15 +1,14 @@
 package au.org.ala.pipelines.transforms;
 
+import static org.junit.jupiter.api.Assertions.*;
+
+import java.util.HashMap;
+import java.util.Map;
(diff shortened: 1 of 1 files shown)


StefanVanDyck:
Damn, seems like I did not include the unit-test I added.
Will recreate and add them once the DEV build is working again.
(Snapshot library dissapeared from the gbif repository?)

vjrj:
> Damn, seems like I did not include the unit-test I added. Will recreate and add them once the DEV build is working again. (Snapshot library dissapeared from the gbif repository?)

Curiously, I was testing it in my side, adding tests and facing the same snapshots errors like you. Anyway the PR LGreatTM.

vjrj (approved):

vjrj:
@StefanVanDyck can you do a `mvn spotless:apply`? Thanks

https://builds.gbif.org/blue/organizations/jenkins/pipelines-multibranch/detail/PR-1226/4/pipeline

StefanVanDyck:
Whoops, keep forgetting :)
Should add some kind of git hook or something.
Thanks @vjrj

PR #1228, opened 2025-11-18
Some additional changes, required to be able to run without the sensi…

StefanVanDyck:
For the Flemish biodiversity portal we are not using the sensitive data service.
According to the README, setting includeSensitiveDataChecks to false should allow running without it.
But at present the indexing fails, complaining about the lack of the avro files, generated by the sensitive data step.

With these changes we can remove both the step, as well as the additional micro-service from our setup.

Do let me know if you would like me to do this in another (possibly cleaner?) way.


vjrj (approved):
LGTM. Thanks @StefanVanDyck

PR #1287, opened 2026-02-19
The read inputstream was never closed correctly.

StefanVanDyck:
Was causing issues with our use of S3 files directly.
The connection pool would very quickly run out, because the were never closed.
Simple fix by using try-with-resources, diff looks scarier than it actually is.


vjrj:
@StefanVanDyck  your friend "spotless" is complaining...

PR #1288, opened 2026-02-20
Add S3 Support

StefanVanDyck:
Adds support for S3 storage by using the "s3://" uri prefix.
Check the example s3 config included in this PR.

Very useful when using AWS EMR (Or any other cloud-provided ephemeral spark cluster).
No need to sync any FS with HDFS every time and very good scaling IO performance.

Might even open the way to using serverless EMR, except the Interpretation step still relies on having a bunch of large Shape File present on the local filesystem. Could Try and add support for loading the shapefilles from S3 directly too.
(Using apache beam FS utilities) But this already makes things much simpler.

BTW, s3 support also work when running locally (given you setup the AWS credentials) or in a non EMR/Spark batch job.
So again no need to sync data between different filestorage systems. 
And makes it easy to debug by running locally using the actual pipeline data used in S3.


StefanVanDyck:
@patkyn Whoops :facepalm: 
Yeah the apikey was definitely not intentional, thanks for spotting it.
I amended the commit (So it should be removed from git history).
Fortunately it was an old one that was already rotated.

As for the paths / the rest of the config file I added it because I thought it might be instructive in how to use s3 instead of local or hdfs. Similar to the other config files there.

Thank you for having a look :pray:

PR #1349, opened 2026-05-04
Fixed Unit tests dependency on environment variables, not set

Wisienkas:
**Fix unit test network binding for local development**

- Added `SparkTestSession` utility to centralise Spark test session configuration, including explicit bind address settings
- Migrated all test classes to use `SparkTestSession` instead of inline `SparkSession.builder()` calls
- Added explicit loopback bind configuration to `HbaseServer` mini-cluster setup
- Added autogenerated Avro schema files to `.gitignore`


djtfmartin (approved):
LGTM

PR #1351, opened 2026-05-05
Nucleotide dna

marcos-lg:
These PRs need to be merged before:
- dwc-api: https://github.com/gbif/dwc-api/pull/20
- gbif-api: https://github.com/gbif/gbif-api/pull/203

To be deployed it needs the `target_gene` vocab to be added to the config.

The `dnaSequenceID` field is kept for backwards compatibility, hence it's not added to the new table but remains in the occurrence table.


pom.xml
```diff
@@ -101,18 +101,19 @@
     <!-- GBIF libraries -->
     <gbif-parsers.version>0.68</gbif-parsers.version>
     <dwca-io.version>3.0.1</dwca-io.version>
-    <gbif-api.version>2.4.3</gbif-api.version>
+    <gbif-api.version>2.4.5-dna-SNAPSHOT</gbif-api.version>
     <gbif-common.version>0.60</gbif-common.version>
-    <dwc-api.version>2.1.10</dwc-api.version>
+    <dwc-api.version>2.1.11-dna-SNAPSHOT</dwc-api.version>
```
djtfmartin: do we want to release gbif-api  + dwc-api before merging ?

marcos-lg: I'll merge them before the PR in pipelines so it compiles but I think I'll release it when we'll be about to release pipelines. I think the only change that can be problematic is the one we saw in the dwca converter that was creating the DNA extension row type in camelCase

PR #1418, opened 2026-06-29
484 index dataset category

ahakanzn:


jhnwllr:
@peterdesmet this might be of interest to you. Soon dataset category will be tied to occurrence records, so that filtering by category will be possible.

PR #1439, opened 2026-07-14
modified ChecklistValidator to use a call to the clb api

marcos-lg:
When merged it can be tested with the rest of the changes in the validator in whaterver env is deployed.


gbif/validator/validator-checklists/src/main/java/org/gbif/pipelines/validator/checklists/ChecklistValidator.java
```diff
@@ -1,165 +1,209 @@
...
   public List<Metrics.FileInfo> evaluate(Path archivePath) throws IOException {
-    Archive archive = DwcFiles.fromLocation(archivePath);
     List<Metrics.FileInfo> results = new ArrayList<>();
-    ArchiveFile core = archive.getCore();
-    if (core.getRowType() == DwcTerm.Taxon) {
-      results.addAll(validateArchive(archive));
+
+    ValidatorResponse validatorResponse =
+        checklistbankWsClient.validateArchive(Files.readAllBytes(archivePath));
+    int datasetKey = validatorResponse.getKey();
```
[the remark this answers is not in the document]
marcos-lg: Done. Added the file name to the path

mike-podolskiy90 (approved):

gbif/validator/validator-checklists/src/main/java/org/gbif/pipelines/validator/checklists/ChecklistValidator.java
```diff
@@ -1,165 +1,209 @@
...
+
+    if (datasetKey == 0) {
+      throw new IllegalStateException("Validation failed with key zero for " + archivePath);
```
mdoering, line 73: The validator requires the clb importer component to be active. On deploys and maintenance this is off and we might want to handle that with a more informative response. you will see a 503 response in that case

gbif/validator/validator-checklists/src/main/java/org/gbif/pipelines/validator/checklists/ChecklistValidator.java
```diff
@@ -1,165 +1,209 @@
...
+          terms.entrySet().stream()
+              .map(
+                  e ->
+                      Metrics.TermInfo.builder()
+                          .term(e.getKey().qualifiedName())
+                          .rawIndexed(e.getValue())
+                          .build())
+              .toList();
+
+      if (rowType == DwcTerm.Taxon) {
```
mdoering, line 97: That is only for dwca. We have also coldp datasets registered. Maybe check that the row type is not a col: prefixed one? There is a ColdpTerm enum. Otherwise all coldp rowtypes  end up as extensions and there is no core

gbif/validator/validator-checklists/src/main/java/org/gbif/pipelines/validator/checklists/ChecklistValidator.java
```diff
@@ -1,165 +1,209 @@
...
-public class ChecklistValidator implements Closeable {
+public class ChecklistValidator {
+
+  private static final int SAMPLE_ISSUES_SIZE = 5;
 
-  @Data
-  @Builder
-  public static class Configuration {
+  // values to wait for validation api response
+  private static final int MAX_WAIT_SECONDS = 720;
```
mdoering, line 34: Thats 12 minutes. By far not enough for large datasets. They can take hours. A COL release would be sth like 6-8h. And it is queued, so there might be considerable waiting time. 

I would do at least 6 better 12 hours. Alternatively I can add a callback to the API that is called when the validation is ready.

mdoering (approved):
Will not work with large lists because the waiting time is far too restrictive. 

See https://github.com/CatalogueOfLife/backend/issues/1552

gbif/validator/validator-checklists/src/main/java/org/gbif/pipelines/validator/checklists/ChecklistValidator.java
marcos-lg: @mdoering so for a coldp dataset what would it be the rowType of the core?

gbif/validator/validator-checklists/src/main/java/org/gbif/pipelines/validator/checklists/ChecklistValidator.java
marcos-lg: @mdoering The callback sounds good. I can increase the waiting time but 6-12 hours sounds like too much because that means the cli will have that thread blocked all that time and we might have issues.

gbif/validator/validator-checklists/src/main/java/org/gbif/pipelines/validator/checklists/ChecklistValidator.java
mdoering: coldp is not a star schema, so "core" is a bit misleading. But the "taxon core" can either be `col:NameUsage` OR `col:`Taxon & `col:Synonym` separately. Note the 2 flavors here: https://github.com/CatalogueOfLife/coldp/blob/master/README.md#schema

gbif/validator/validator-checklists/src/main/java/org/gbif/pipelines/validator/checklists/ChecklistValidator.java
mdoering: if the imports are down e.g. because of maintenance it can be 1-2 days even for a small dataset.
Lets work towards the callback - you still need to deal with waiting for up to 1-2 days...

mdoering:
CLB commit to support callbacks: https://github.com/CatalogueOfLife/backend/commit/25ca7727fd91f4d4798150560b095604479cbd7c

Will go to dev today, but to prod only in late august.

PR #1453, opened 2026-07-24
fix(humboldt): NPE guard on getGeospatialScopeAreaUnit()

Wisienkas:
Ingestion where it was null yielded a hard failure instead of a graceful return
```
Caused by: java.lang.NullPointerException 
org.gbif.pipelines.core.interpreters.extension.HumboldtInterpreter.lambda$checkAreas$4(HumboldtInterpreter.java:293)
```

### DIFF
--- a/sdks/core/src/main/java/org/gbif/pipelines/core/interpreters/extension/HumboldtInterpreter.java
+++ b/sdks/core/src/main/java/org/gbif/pipelines/core/interpreters/extension/HumboldtInterpreter.java
@@ -282,16 +282,23 @@ public class HumboldtInterpreter {
     };
   }
 
-  private static Function<Humboldt, List<String>> checkAreas() {
+  static Function<Humboldt, List<String>> checkAreas() {
     return humboldt -> {
-      if (humboldt.getGeospatialScopeAreaValue() != null
-          && humboldt.getTotalAreaSampledValue() != null
-          // we only check them if they have the same unit since we can't interpret the units until
-          // we have a vocabulary
-          && humboldt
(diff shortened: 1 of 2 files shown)


sdks/core/src/main/java/org/gbif/pipelines/core/interpreters/extension/HumboldtInterpreter.java
```diff
@@ -282,16 +282,22 @@ private static Function<Humboldt, List<String>> checkTargetExclusions() {
...
-          && humboldt
-              .getGeospatialScopeAreaUnit()
-              .equalsIgnoreCase(humboldt.getTotalAreaSampledUnit())
-          && humboldt.getGeospatialScopeAreaValue() < humboldt.getTotalAreaSampledValue()) {
+      Double geospatialScopeAreaValue = humboldt.getGeospatialScopeAreaValue();
+      Double totalAreaSampledValue = humboldt.getTotalAreaSampledValue();
+      String geospatialScopeAreaUnit = humboldt.getGeospatialScopeAreaUnit();
+      String totalAreaSampledUnit = humboldt.getTotalAreaSampledUnit();
+
+      boolean unitsComparable =
```
Wisienkas, line 293: Replaced comment with an intent driven intermediate variable name

marcos-lg: the variable name is missing the part of the comment that says we do this because we don't have a vocabulary yet

marcos-lg (commented):
Only the comment, then feel free to merge it.

PR #1489, opened 2026-08-20
[Feature] DWCDP Mapping engine - Ingest

Wisienkas:
Here's the trimmed version:

---

### Summary

This PR replaces the previous hand-written, imperative DwC-DP → DwC-A conversion with a new **declarative, config-driven mapping engine** under `org.gbif.pipelines.spark.dwcdp.mapping`.

Instead of each DwC-A target being produced by bespoke Spark join code scattered across builder classes, mappings are now **declared** (schema path + cardinality + target semantics), **compiled** and **validated** ahead of execution, and only then **executed** against Spark datasets by a small set of generic executors. This is a full cutover — the legacy builder package is deleted, not kept alongside the new engine.

### Why

The old builder-per-extension approach required hand-rolled join logic for every new field/extension, made ambiguous relationships (one-to-many, optional vs. required) implicit in Spark code, and made it hard to reason about what was and wasn't covered without reading Spark execution code. The new engine centralizes schema navigation, cardinality, and target-aggregation semantics so that:
- Mapping mistakes fail fast at **validation/compilation time**, not deep inside a Spark job.
- Adding/changing a mapping is a config change, not new join code.
- The same target semantics apply consistently whether the target lands on a DwC-A core or an extension row.

### What's new — architecture

Four separated layers, each with its own package: **schema** (DwC-DP resources, relations, path/graph resolution), **definition** (the declarative mapping DSL), **config** (Event/Occurrence/domain mapping declarations), **compilation** (validation, resolution, pruning, diagnostics), and **execution** (the Spark implementation of compiled semantics), tied together by a top-level `DwcDpMappingEngine`.

Key design points (documented in full in the new `dwcdp-to-dwca-mapping.md`):
- **Schema-backed navigation** — mappings walk the schema graph instead of hard-coding joins.
- **Explicit ambiguity handling** — cardinality (fan-out vs. exactly-one) and optional/required relations must be declared; an exactly-one path with zero or multiple matches deliberately contributes nothing rather than picking arbitrarily.
- **Structural vs. DwC identity** — DwC-DP structural primary/foreign keys drive navigation and row identity but are never emitted as DwC terms; a missing natural `eventID`/`occurrenceID` gets a deterministic `urn:gbif:dwcdp:...` fallback instead of a copied structural key.
- **Shared target semantics** — the same compiled target producer behaves identically whether it lands on the core or an extension row.
- **Deliberate, documented lossiness** — e.g. Event-core media promotion for Occurrence/Material-owned Multimedia, deterministic list flattening/dedup, taxonomic fallback via a single resolved row. See §15 of the doc.
- **Dataset pruning** — queries are pruned based on the actual `datapackage.json` and downstream mapping requirements, avoiding unnecessary reads/joins for resources not present in a given package.
- **Path-prefix caching** — shared navigation prefixes are reused across mapping fragments as a pure execution optimization.

### Coverage carried over / extended in the new engine

All domains previously handled by the builder pipeline are re-implemented as declarative config, plus additions made during the migration: Agent/AgentRole, Assertion, Humboldt, Identification (incl. DNA/eDNA), Identifier, Multimedia, Nucleotide, Reference/bibliographic resources, Chronometric, generalized Protocol references, Material (incl. geological context, provenance, usage policy, media, assertion, identifier), Event/Occurrence enrichment, and virtual Occurrence creation for Event materials without a local evidence Occurrence.

### Other changes bundled in this PR

- The verbatim converter is rewritten as an orchestrator around the new engine, with an expanded, more detailed conversion report writer.
- New `dwcdpDetailedConversionReport` config flag (off by default) — when enabled, computes per-domain join funnels (resolved/ambiguous/unresolved breakdowns) for the conversion report. Off by default since each funnel is an extra Spark pass, meant for investigating a specific dataset/attempt rather than routine runs.
- The occurrence- vs. event-only workflow routing decision now uses the actual computed occurrence count rather than an upstream message flag, with added logging for the routing decision and missing metrics.
- New dependency on `dwc-dp-schemas`, backing the schema graph/loader.
- New extensive documentation: `gbif/ingestion/spark-jobs/dwcdp-to-dwca-mapping.md`, covering architecture, design principles, the mapping DSL, compilation/validation, Spark execution, core/extension merges, identity rules, lossy-conversion policy, and a step-by-step guide for adding/changing a mapping.

### Removed

The entire legacy builder package (core builders, extension/join builders, and their term-resolution helpers) is deleted, along with its corresponding unit tests.

### Testing

New unit and Spark-integration test coverage across every new layer — schema/relation resolution, mapping-path/plan composition, compiler behavior (including precedence and dataset-pruning reachability), each domain config class, and execution semantics (extension materialization, target expressions, path execution, core/extension record assembly) — plus an updated end-to-end verbatim converter test and updated coordinator routing/config tests.

### Notes for reviewers

The most useful review entry points mirror the doc's "Implementation entry points" section: start from `EventDwcaMapping.current(...)` / `OccurrenceDwcaMapping.current(...)` for *what* is mapped, `MappingCompiler`/`MappingValidator` for *what's legal*, and the Spark executor classes for *how it runs on Spark*.


sdks/config/src/main/java/org/gbif/pipelines/core/config/model/PipelinesConfig.java
```diff
@@ -28,6 +28,16 @@ public class PipelinesConfig implements Serializable {
...
 
+  /**
+   * When {@code true}, {@code DwcDpVerbatimConverter} additionally computes per-builder join
+   * funnels (resolved/ambiguous/unresolved breakdowns for agent, protocol, provenance,
+   * geological-context, organism, etc. resolution) and includes them in {@code
+   * conversion-report.txt}. Off by default — each funnel is an extra Spark pass over the relevant
+   * source table purely for reporting, so this is meant to be switched on for a specific
+   * dataset/attempt when investigating a conversion, not left on in normal runs.
+   */
+  private boolean dwcdpDetailedConversionReport = false;
```
Wisienkas, line 39: can be removed again

PR #1495, opened 2026-08-24
Fix/dwcdp occurrence media

Wisienkas:
- Adding OccurrenceID to Media from the occurrence-media resource bridge
- fixed fallback ids for eventID on occurrence


fmendezh (changes requested):
let's convert this into a generic solution that adds the occurrenceID for all the extensions coming from tables linked to an occurrence table in DP

